PR #3, opened 2019-05-22
Add Devfiles which are based on Che 7 Stacks

sleshchenko:
### What does this PR do?
This PR adds Devfiles which are based on Che 7 Stacks listed in https://github.com/eclipse/che/issues/12620

Also, devfiles structure is changed and now a registry contains two files `meta.yaml` (with displayName, description, icon) and `devfile.yaml`.

devfile-registry

https://github.com/eclipse/che/issues/13353


devfiles/java-mysql-theia-kubernetes.yaml
```diff
@@ -0,0 +1,113 @@
...
+name: java-mysql-theia-kubernetes
+components:
+  -
+    type: cheEditor
+    id: eclipse/che-theia/next
+  -
+    type: chePlugin
+    id: eclipse/che-machine-exec-plugin/0.0.1
+  -
+    type: dockerimage
```
sleshchenko: I used `dockerimage` component because it is simpler than `kubernetes`, but original Che 7 stack contains kubernetes list (actually because there was no an ability to use any other recipe type because of multiple dockerimages).

With `kubernetes` component it would look like the following:
```yaml
  -
    type: kubernetes
    alias: web
    referenceContent: |
      kind: List
      items:
      -
        apiVersion: v1
        kind: Pod
        metadata:
          name: web
        spec:
          containers:
          -
            image: registry.centos.org/che-stacks/centos-jdk8
            name: dev
            ports:
            - containerPort: 8080
              protocol: TCP
            resources:
              limits:
                memory: 512Mi
            volumeMounts:
            - name: projects
              mountPath: /projects
          volumes:
          - name: projects
            persistentVolumeClaim:
              claimName: projects
      -
        apiVersion: v1
        kind: PersistentVolumeClaim
        metadata:
          name: projects
          labels:
            app.kubernetes.io/name: employee-manager
            app.kubernetes.io/component: nodejs-app
        spec:
          accessModes:
           - ReadWriteOnce
          resources:
            requests:
              storage: 1Gi
```
vs dockerimage
```yaml
    type: dockerimage
    alias: web
    dockerimage: registry.centos.org/che-stacks/centos-jdk8
    memoryLimit: 512Mi
    mountSources: true
    endpoints:
      - name: tomcat8
        port: 8080
````
Just wanted to ask if everybody is OK to use `dockerimage` instead of `kubernetes` whether it is possible.
@skabashnyuk @l0rd WDYT?

Comment (sleshchenko) [reply devfiles/java-mysql-theia-kubernetes.yaml]: For this particular case, it makes sense to leave kubernetes components since it's named as `java-mysql-theia-kubernetes`, but in other devfiles, I'm going to use dockerimage where it is possible to not `kuberentes` is mentioned in description or name.

File: devfiles/apache-camel-springboot-che7.yaml [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+displayName: Apache Camel based projects on Che 7
+description: Stack with environment ready to develop Integration projects with Apache Camel based on SpringBoot.
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+specVersion: 0.0.1
+name: apache-camel-springboot-che7
+components:
+  -
+    type: cheEditor
+    id: eclipse/che-theia/next
+    memoryLimit: 512Mi
```
l0rd: Can we avoid adding theia and the exec plugin?

sleshchenko: We can but I'm not sure if we should.

I mean, as about exec plugin - I guess we can and should remove it from Devfiles, will do it.
But about Theia, it would make Devfiles smaller (- 3 lines) but at the same time:
1. Memory limit could not be overridden for theia in such case (not sure if it's needed, just did the same as Che 7 Stack has).
2. Devfile contains plugins which are not compatible with any other editor except Theia. Just theoretical case: what if Che Server is configured with `gwt`/`jupyter` editor, and devfile contains plugins for Theia but Theia is not declared.
@l0rd Do you think we should remove theia as well?

devfiles/apache-camel-springboot-che7.yaml
```diff
@@ -0,0 +1,56 @@
+---
+displayName: Apache Camel based projects on Che 7
```
l0rd: @sleshchenko @slemeur @skabashnyuk I am coming late to the discussion and there may be some details that you have already discussed but I would like to propose a slightly different structure/format for the registry. 

The devfiles in this registry are going to be the reference devfiles for any developer who wants to create her own devfile. She will just copy/paste one of these devfile and customize it. But a couple of things are not *standard* in current structure:

- `displayName`, `description` and `icon` are not attributes of a standard devfile 
- the standard devfile name is `devfile.yaml` 

A possible solution would be to have the following structure for a devfile:

```
devfiles
   | 
    ---  apache-camel-springboot   # No mention of che7 here
              |
              |--- meta.yaml       # Contains the stack details as displayName, icon and descriptin
              ' --- devfile.yaml   # A standard devfile
```

devfiles/apache-camel-springboot-che7.yaml
l0rd: @sleshchenko yes I believe we should remove it. The day we will have real world `gwt/jupyter` use cases we will think about it (there will be probably a a dedicated devfile registry).

sleshchenko: OK will do it for stacks where memory limit for theia is not needed, will ask authors

devfiles/che7-preview-plugin-dev.yaml
```diff
@@ -0,0 +1,26 @@
+---
+displayName: Che 7 Dev
+description: Workspace.next sidecars and Theia as IDE with the tooling for plugin development
```
sleshchenko: This stack looks like sample, as far as I remember it was used on EclipseCon 2018.
@l0rd Do you have an idea whether we still need since we have another stack for Theia Development contributed and used by IDE2 team https://github.com/eclipse/che-devfile-registry/pull/3/files#diff-c73cf9a15f50d1a8ca46c423fe23abf3R3 ?

devfiles/apache-camel-springboot-che7.yaml
sleshchenko: makes sense to me.

Just a small question/proposal: can icon field be a relative link to icon file that is placed in `"stack"` directory?

devfiles/che7-preview-plugin-dev.yaml
l0rd: Please refer to https://github.com/eclipse/che/issues/12620. We need only the stacks listed in that issue.

devfiles/apache-camel-springboot-che7.yaml
sleshchenko: One more question related to `meta.yaml` of stack, during converting stacks to devfiles, I removed `tags` and `component` fields:
```
  tags: ["ws.next", "Theia", "Java"]
  components:
  - name: Centos
    version: '7'
  - name: OpenJDK
    version: 1.8.0_181
  - name: NodeJS
    version: 8.12.0
  - name: NPM
    version: 6.4.1
```
Should they be in a `meta.yaml`? In the same format or different?
Note that having `components` field in `meta.yaml` and `devfile.yaml` may be confusing, but for me, it looks clear.

l0rd: Yes I mean for the icon I would make it even simpler: use a standard name (e.g. logo.svg) so that you don't even need to add it to the meta.yaml.

sleshchenko: And also about `publisher`, for stacks we used to have `publisher` specified.
I think we can just drop it out, but wanted to get confirmation that we really do not need it.
Otherwise, folder structure may be more similar to that we have for Plugin Registry

sleshchenko: >
>  ```
>  devfiles
>    | 
>     ---  apache-camel-springboot   # No mention of che7 here
>               |
>               |--- meta.yaml       # Contains the stack details as displayName, icon and descriptin
>               ' --- devfile.yaml   # A standard devfile
>  ```

Just wanted to leave here alternative way not to mix `devfile` and `meta` fields. Just the same folder structure that we have now, but object is like the following:
```yaml
displayName: ----
description: 
icon:
devfile:
  projects:
  components:
  ...
```
I agree with Mario that from reusing devfiles (copy/paste) parspective, a separate file with devfile looks more convenient.

devfiles/che7-preview-plugin-dev.yaml
sleshchenko: Thanks for sharing. Have not noticed this issue before

devfiles/apache-camel-springboot-che7.yaml
l0rd: Let's leave legacy `components` attribute out. You can add `tags` and `publisher` to meta.yaml (or do that in a separate PR).

devfiles/che7-preview-plugin-dev.yaml
sleshchenko: Checked https://github.com/eclipse/che/issues/12620 and only these stacks are presented there
- apache-camel-springboot-che7.yaml
- dotnet.yaml
- java-gradle.yaml
- go.yaml
- java-maven.yaml
- python.yaml

Also there is mentioned Che-Theia Developer stack, but referenced is sunix's Devfile but not Che 7 stack that we have. So, the following stacks will be dropped out.
- che7-development-che-theia-and-theia-plugins.yaml
- che7-preview-plugin-dev.yaml
- java-mysql-theia-kubernetes.yaml
- che7-preview.yaml

devfiles/apache-camel-springboot-che7.yaml
sleshchenko: @l0rd In this issue `512m` is recommended to use for Theia https://github.com/eclipse/che/issues/12620
But default value in Plugin Registry is `1536M` https://che-plugin-registry.openshift.io/v3/plugins/eclipse/che-theia/next/

Should I create an issue to make default memory limit for Theia 512, 
or keep theia component in these devfiles to set 512 memory limit?

l0rd: Yes it should be 512m by default. Don't know why we still have the old value but it must be changes.

sleshchenko: The corresponding PR is created https://github.com/eclipse/che-plugin-registry/pull/156

l0rd:
@sleshchenko as a consequence of https://github.com/eclipse/che/pull/13414 we should now use `latest` as versions for plugins.

devfiles/apache-camel-springboot-che7.yaml
slemeur: ok for me

sleshchenko: Changed to 

> ```
> devfiles
>   | 
>    ---  apache-camel-springboot   # No mention of che7 here
>              |
>              |--- meta.yaml       # Contains the stack details as displayName, icon and descriptin
>              ' --- devfile.yaml   # A standard devfile
> ```

sleshchenko:
@l0rd 

> @sleshchenko as a consequence of eclipse/che#13414 we should now use latest as versions for plugins.

PR you mentioned is not merged yet. Stack description declares its components version, like (Java Stack with OpenJDK 11 and Gradle 5.2.1), are you sure that plugins should use `latest` since java version is fixed? Is the latest java plugin compatible with OpenJDK 11?
BTW Are you OK to create a separate issue for that?

l0rd:
@sleshchenko yes I will create a separate issue so we can discuss that. No need to include that in this PR.

l0rd:
FYI https://github.com/eclipse/che/issues/13426

PR #4, opened 2019-05-27
Add tags to meta.yaml

sleshchenko:
### What does this PR do?
This PR follows the following discussion https://github.com/eclipse/che-devfile-registry/pull/3#discussion_r287310042
and it adds tags to meta.yaml.
Values are copied/pasted from existing Che 7 stacks from Che Repository.

https://github.com/eclipse/che/issues/13353

### DIFF
--- a/devfiles/apache-camel-springboot-che7/meta.yaml
+++ b/devfiles/apache-camel-springboot-che7/meta.yaml
@@ -1,4 +1,5 @@
 ---
 displayName: Apache Camel based projects on Che 7
 description: Stack with environment ready to develop Integration projects with Apache Camel based on SpringBoot.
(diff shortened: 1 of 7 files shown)


sleshchenko:
cc @ashumilova

ashumilova:
@slemeur this PR adds the tags to devfile's metadata for stacks, which we use on UI to bind with project templates (our samples). Just rechecking - we keep the sample projects (which depend on choosen devfile) on workspace creation page, right?

l0rd:
@ashumilova @slemeur I would rather embed the project in the devfile itself. There have been some discussions about how hard it can be to maintain and sync stacks to support n samples. Commands may work on a sample but not having sense for other samples. And we can have some weird associations. Having everything defined at the stack level instead (projects + components + commands) is easier to maintain and easier for a user to choose (he only needs to select the stack).

PR #8, opened 2019-06-11
Update the devfiles to the latest version of the devfile schema.

metlos:
### What does this PR do?
`$TITLE`

https://github.com/eclipse/che/issues/13336


sleshchenko:
Note that it will be needed to update Devfile Registry on OpenShift.io when new docker image will be build after merging of this PR.
Docker image should be built here https://ci.centos.org/job/devtools-che-devfile-registry-build-master/
New has should be set here https://github.com/openshiftio/saas-openshiftio/blob/master/dsaas-services/che-devfile-registry.yaml#L2

skabashnyuk:
I can do that.

PR #10, opened 2019-06-12
Use latest versions of plugins in devfiles

mshaposhnik:
Make devfiles in regisrtry use latest versions of plugins


sleshchenko:
Looks liek it's related to https://github.com/eclipse/che/issues/13426

PR #13, opened 2019-06-17
Adds an example project to go devfile.

metlos:
### What does this PR do?

This adds an example project to the Go devfile in the registry.

Note that contrary to the requirement in https://github.com/eclipse/che/issues/13529 this PR does NOT add the https://github.com/xesina/golang-echo-realworld-example-app project to the devfile, because that project uses Go modules, available only in Go 1.11, while our che plugin for Go uses Go 1.10.7. 

Note also that I made some slight adjustments to the plugin definition and image to ever so slightly improve the experience with them.

eclipse/che-plugin-registry#165
https://github.com/eclipse/che-theia/pull/288


sleshchenko:
@metlos It would be good to see some commands, like the following https://github.com/eclipse/che/blob/master/ide/che-core-ide-templates/src/main/resources/samples.json#L269

metlos:
> @metlos It would be good to see some commands, like the following https://github.com/eclipse/che/blob/master/ide/che-core-ide-templates/src/main/resources/samples.json#L269

Completely forgot about this, but now an example command is included.

PR #14, opened 2019-06-18
Update gradle mounted volume path to match dockerfile

amisevsk:
### What does this PR do?
Modifies the volume used for the `.gradle` directory to be `/home/gradle/.gradle`. This is necessary because the [dockerfile](https://github.com/keeganwitt/docker-gradle/blob/7c606d205bcf13551f8203e12174c37004383694/jdk8/Dockerfile#L17) used specifies this as a `VOLUME`, which causes issues when that path is not mounted via PVC.

I have tested this devfile and it appears to work (at least, as well as community docker images generally do -- see https://github.com/eclipse/che/issues/13454) but further testing is appreciated.

https://github.com/eclipse/che/issues/13384, see [comment](https://github.com/eclipse/che/issues/13384#issuecomment-503175239) for more explanation.

### DIFF
--- a/devfiles/java-gradle/devfile.yaml
+++ b/devfiles/java-gradle/devfile.yaml
@@ -20,7 +20,7 @@ components:
     args: ['infinity']
     env:
       - name: GRADLE_USER_HOME
-        value: /home/user/.gradle
+        value: /home/gradle/.gradle
       - name: JAVA_OPTS
         value: "-XX:MaxRAMPercentage=50 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
           -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90
@@ -32,14 +32,14 @@ components:
       - name: PS1
         value: "$(echo ${0})\\$ "
       - name: HOME
-        value: /home/user
+        value: /home/gradle
     memoryLimit: 512Mi
     endpoints:
       - name: '8080/tcp'
         port: 8080
     volumes:
       - name: gradle
-        containerPath: /home/user/.gradle
+        containerPath: /home/gradle/.gradle
     mountSources: true
 commands:
   -


devfiles/java-gradle/devfile.yaml
```diff
@@ -20,7 +20,7 @@ components:
     args: ['infinity']
     env:
       - name: GRADLE_USER_HOME
-        value: /home/user/.gradle
+        value: /home/gradle/.gradle
```
sleshchenko: just a question: do we need two homes? or 
```
      - name: HOME
        value: /home/user
```
Should be changed to 
```
      - name: HOME
        value: /home/gradle
```
?

l0rd: I set to `HOME` to `/home/user` originally because that's how legacy Che workspaces worked and there were still some environment variables injected by wsmaster that assumed that. I don't know if that's still the case.

`/home/gradle/.gradle` is persisted as a volume right? data should not be wiped out. 

I am afraid that setting `/projects/` as `HOME` may have unexpected side effects. We may want to set the [k8s container workingDir](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#container-v1-core) instead.

amisevsk: @l0rd That makes sense to me but I'm still not sure the best way to proceed here. While `/home/gradle/.gradle` is persisted, `/home/gradle` is not; are we okay with this (e.g. it would prevent using a persistent `.bashrc`, etc.). This issue is probably another, larger discussion, though. 

I've updated `$HOME` to be `/home/gradle`. What is currently happening, AFAICT, is that whatever directory we're calling `$HOME` is being created at container start up, so I don't think this will cause issues.

l0rd: We are ok not to persist full `$HOME`. This is a cool idea (a user would find his dotfiles in every workspace) but I don't think there has ever been a requirement for that.

l0rd: Anyway I am +1 to merge this PR as it is

PR #15, opened 2019-06-18
Add a nodejs stack

l0rd:
### What does this PR do? 

Add a new nodjes stack with an hello-world express web application

https://github.com/eclipse/che/issues/12620

### DIFF
--- a/devfiles/nodejs/devfile.yaml
+++ b/devfiles/nodejs/devfile.yaml
@@ -1,0 +1,38 @@
+---
+apiVersion: 1.0.0
+metadata:
+  name: nodejs
+projects:
+  -
+    name: nodejs-web-app
+    source:
+      type: git
+      location: "https://github.com/che-samples/web-nodejs-sample.git"
+components:
+  -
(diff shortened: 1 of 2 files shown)


sleshchenko:
@l0rd If you plan to add more devfiles, please let us know here https://github.com/eclipse/che/issues/13529#issuecomment-501746114

ibuziuk:
Works like a charm on che.openshift.io - 
image

I have added this devfile into the list of replacement for `getting-started` - https://github.com/ibuziuk/my-che-devfiles/tree/master/nodejs-quickstart

Also, once https://github.com/eclipse/che/pull/13561 we could add `supportArbitraryUser: 'true'` (not sure but I think that `/etc/passwd` should be writable)

l0rd:
@sleshchenko @metlos I've created this PR on the nodejs stack because it was assigned to me (c.f. #1620) and #13529 was more about the sample rather than the stack itself. But I am happy to let you work on the rest. I have just assigned the react stack to myself because I had started working on it.

PR #20, opened 2019-06-24
Moved working directory to a dedicated field instead of using 'cd' command

sleshchenko:
### What does this PR do?
This PR moves the working directory to a dedicated field instead of using 'cd' command.
It makes commands shorter and obvious.

Devfiles are not tested after changes but it should not break anything.

N/A


sleshchenko:
There is one more improvement that can be done - use `${current.project.path}` macros, but the reason why I did not do it in this PR - I'm not sure that should be a default behavior and that it's still working

PR #22, opened 2019-06-28
Add a mongo db application stack.

metlos:
### What does this PR do?
Adds a stack with nodejs and mongodb with an example application.

Current problems:
* the application started through the command seems to stop after a couple of minutes (this seems to be an issue with Che or even K8s in general?)
* The example curl command to create a user is stuck and generates no response. I didn't find the root cause for this yet.

### DIFF
--- a/devfiles/nodejs-mongo/meta.yaml
+++ b/devfiles/nodejs-mongo/meta.yaml
@@ -1,0 +1,6 @@
+---
+displayName: NodeJS RealWorld Web Application
+description: Default stack with NodeJS 10
+tags: ["NodeJS", "Express", "RealWorld", "Theia", "ubi8"]
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+globalMemoryLimit: 1686Mi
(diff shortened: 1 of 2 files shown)


devfiles/nodejs-mongo/meta.yaml
```diff
@@ -0,0 +1,6 @@
+---
+displayName: NodeJS RealWorld Web Application
```
l0rd: I would rather call it "NodeJS MongoDB Web Application"

l0rd:
> the application started through the command seems to stop after a couple of minutes (this seems to be an issue with Che or even K8s in general?)

maybe related to #13620?

metlos:
> the application started through the command seems to stop after a couple of minutes (this seems to be an issue with Che or even K8s in general?)

This is caused by https://github.com/eclipse/che/issues/13620 and so can be dealt with separately.

> The example curl command to create a user is stuck and generates no response. I didn't find the root cause for this yet.

This was caused by https://github.com/eclipse/che/issues/8134 and thus was an issue with my local setup.

So this PR is ready for review I think.

PR #23, opened 2019-07-02
Include vscode Git plugin to all dev files

vinokurig:
### What does this PR do?
Include vscode Git plugin to all dev files

https://github.com/theia-ide/theia/issues/4266
Is needed for https://github.com/eclipse/che/issues/11867


benoitf:
Hello, I'm thinking as git is quite a "default" extension, it should be included in the default che-theia image and not be part of the registry

PR #27, opened 2019-07-03
#14076 add ubi8 based dockerfile and update shell scripts to not require find & yq so they're more portable

nickboldt:
* replace need for yq with shell processing (cat|grep|sed); add new Dockerfile based on ubi8 (single stage = faster build); make variable names more consistent between scripts.

Change-Id: I7e4e84c8c75f3ec89e8dfbf066fa797f938d7592


skabashnyuk:
@nickboldt I have some concerns that add 5Mb binary to Eclipse Github repository is a  good idea (IP, it's not a source code, it's not a devfile). Is there any way to reference this binary from somewhere else?

benoitf:
I had the same remark but review comment was pending

Dockerfile.ubi8
```diff
@@ -0,0 +1,38 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+FROM registry.redhat.io/ubi8-minimal as builder
+USER 0
+WORKDIR /build/
+
+COPY yq_linux_amd64_2.4.0 /usr/bin/yq
```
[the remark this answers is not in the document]
nickboldt: I need to investigate why the existing Dockerfiles require yq. If we can bypass that, then we don't need a builder image which includes yq (not RH supported, not in Brew/Pulp)

nickboldt:
This is ready for re-review and merge into 7.1 when you guys are ready for it. Not urgent for 7.0 but we WILL need a ubi8 based solution in CRW 2 and if we can have it upstream in Che 7.1 that's one less fork for me to maintain downstream. :)

Mario said:

> I am +1 with replacing `yq` but that’s a nice to have rather than an hard requirement

index.sh
```diff
@@ -12,31 +12,35 @@ set -e
...
 
     FIRST_LINE=true
     echo "["
 
-    ## now loop through meta.yaml files
-    for i in "${arr[@]}"
-    do
-        if [ "$FIRST_LINE" = true ] ; then
+    ## loop through meta.yaml files
+    for i in $(ls -1 ${devfilesDir}/*/meta.yaml | sort); do
```
benoitf: It seems shellcheck is reporting some warnings

nickboldt: can you tell me how/where to run shellcheck and what the warning(s) are?

benoitf: copy paste into https://www.shellcheck.net/

And there is a fixme button

nickboldt: ok, I can do some replacements for cat | grep --> grep and wrap some quotes around stuff.

But replacing `ls ` with `find` means the container needs to install `find`, and assuming that the devfiles folders DON'T contain spaces, there's no need to use find here.

skabashnyuk:
I have a couple of additional questions?
- What problem is this pr is solving? "replace the need for yq with shell processing (cat|grep|sed)"? yq exists only in build image and replace it with (cat|grep|sed) doesn't look much easier to read and understand.
- What will happen with the original Dockerfile?

nickboldt:
> What problem is this pr is solving? "replace the need for yq with shell processing (cat|grep|sed)"? yq exists only in build image and replace it with (cat|grep|sed) doesn't look much easier to read and understand.

It solves the problem that che devfile registry depends on an intermediate container that doesn't exist inside Brew, so I will need to fork this project's build in order to productize it as CRW 2.

If you don't depend on something I can't use, then upstream and downstream become more closely aligned.

> What will happen with the original Dockerfile?

Ideally my ubi8 based on will become your default too. But if you prefer to keep a separate one (as you have for rhel based https://github.com/eclipse/che-devfile-registry/blob/master/Dockerfile.rhel#L18 dockerfile) then that's OK too. Less ideal, but at least the dockerfile can be maintained in upstream instead of being in a completely different repo.

skabashnyuk:
Even it's a legitimate problem for CRW then for the community it might be confusing especially if it makes contribution harder because of support multiple things that are doing the same job.
 Lets review a couple of alternatives:

1. ubi result image. What if result upstream image would be ubi based. (we replace this image https://github.com/eclipse/che-devfile-registry/blob/master/Dockerfile#L18 with ubi) could you reuse it without rebuilding?
2. Could you use `quay.io/eclipse/che-devfile-registry` as a source in multistage build and repackage it for CRW bandle?
3.  What if we build `quay.io/eclipse/che-jq-build` image and reuse it in upstream and downstream?

l0rd:
@nickboldt we are trying to solve an upstream and downstream problem with one unique solution. Implementing and maintaining 2 different solutions is a big waste of time and we should all agree that we want to avoid it. 

@skabashnyuk I am +1 with your 3rd alternative and -1 using UBI based images upstream. You need to be registered to Customer Portal and authenticated to pull images from the Red Hat catalog. And anyway the `httpd` image is not @nickboldt problem. His problem is the `jq` image. 

Anyway @nickboldt I think you need to solve the offline plugin registry as well right? That means that you need `jq` in the plugin registry right?

nickboldt:
> maintaining 2 different solutions is a big waste of time and we should all agree that we want to avoid it.

Yes, that's why I'm pushing for a fix in upstream instead of just forking in downstream. :)

> You need to be registered to Customer Portal and authenticated to pull images from the Red Hat catalog. 

The exception to this rule is **UBI images, which can be fetched from the unauth'd `registry.access.redhat.com`** and the auth'd-only `registry.redhat.io`. That's why we have ubi images in the Che operator, but using the unauth'd repo: https://github.com/eclipse/che-operator/blob/master/pkg/deploy/defaults.go#L42-L43 (for CRW we use reg.rh.io, since you need that registry to get ANY of the CRW images themselves).

> 3. What if we build quay.io/eclipse/che-jq-build image and reuse it in upstream and downstream?

Productization rules require that I can only use a RHEL based image from a RH registry in Brew, so I would need a Dockerfile in this repo (and in fact all the Che repos that need to be productized as container builds) that I can mirror into a pkgs.devel repo to drive a Brew build of that image, and use that image for downstream CRW builds. 

I'd prefer if the image is as close as possible in upstream to downstream, like we do for the che/crw operator container build:

https://github.com/eclipse/che-operator/blob/master/Dockerfile vs. 
http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-operator/tree/Dockerfile?h=crw-2.0-rhel-8

Dockerfile.ubi8
```diff
@@ -0,0 +1,31 @@
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+FROM registry.redhat.io/rhel8/httpd-24:1-30
```
nickboldt: FYI I tried to switch this to use access.registry.redhat.com and was told:

> `$➔ docker pull registry.access.redhat.com/rhel8/httpd-24:1-42`
> Error response from daemon: unsupported: This repo requires terms acceptance and is only available on registry.redhat.io

nickboldt:
Closing. See https://github.com/eclipse/che-devfile-registry/pull/78 for a more atomic cleanup of dockerfiles and scripts.

PR #31, opened 2019-07-11
che #13813: Pushing rhel images to 'quay.io/openshiftio' and centos images to 'quay.io/eclipse-che' during the CI build

ibuziuk:
### What does this PR do?
Pushing rhel images to 'quay.io/openshiftio' and centos images to 'quay.io/eclipse-che' during the CI build

https://github.com/eclipse/che/issues/13813

Should be merged only once dedicated cico job pr is merged - https://github.com/openshiftio/openshiftio-cico-jobs/pull/1035

### DIFF
--- a/cico_build.sh
+++ b/cico_build.sh
@@ -22,6 +22,8 @@ function load_jenkins_vars() {
             DEVSHIFT_TAG_LEN \
             QUAY_USERNAME \
             QUAY_PASSWORD \
+            QUAY_ECLIPSE_CHE_USERNAME \
+            QUAY_ECLIPSE_CHE_PASSWORD \
             JENKINS_URL \
             GIT_BRANCH \
             GIT_COMMIT \
@@ -60,26 +62,30 @@ function deploy() {
 
   if [ "$TARGET" == "rhel" ]; then
     DOCKERFILE="Dockerfile.rhel"
+    ORGANIZATION="openshiftio"
     IMAGE="rhel-che-devfile-registry"
   else
     DOCKERFILE="Dockerfile"
+    ORGANIZATION="eclipse-che"
     IMAGE="che-devfile-registry"
+    # For pushing to quay.io 'eclipse-che' organization we need to use different credentials
+    QUAY_USERNAME=${QUAY_ECLIPSE_CHE_USERNAME}
+    QUAY_PASSWORD=${QUAY_ECLIPSE_CHE_PASSWORD}
   fi
 
   if [ -n "${QUAY_USERNAME}" ] && [ -n "${QUAY_PASSWORD}" ]; then
     docker login -u "${QUAY_USERNAME}" -p "${QUAY_PASSWORD}" "${REGISTRY}"
   else
-    echo "Could not login, missing credentials for the registry"
+    echo "Could not login, missing credentials for pushing to the '${ORGANIZATION}' organization"
   fi
 
-  # Let's deploy
+  # Let's build and push images to 'quay.io'
   docker build -t ${IMAGE} -f ${DOCKERFILE} .
 
   TAG=$(echo "$GIT_COMMIT" | cut -c1-"${DEVSHIFT_TAG_LEN}")
 
-  tag_push "${REGISTRY}/openshiftio/$IMAGE:$TAG"
-  tag_push "${REGISTRY}/openshiftio/$IMAGE:latest"
-  echo 'CICO: Image pushed, ready to update deployed app'
+  tag_push "${REGISTRY}/${ORGANIZATION}/$IMAGE:$TAG"
+  echo "CICO: Images pushed to 'quay.io/openshiftio', ready to update deployed app"
 }
 
 function cico_setup() {


ibuziuk:
The only thing I'm not sure about - are we already pushing *rhel* (not centos) images to `quay.io/openshiftio` - https://quay.io/repository/openshiftio/che-devfile-registry?tab=tags ?
the image name is `che-devfile-registry` and I would expect that rhel image name would be `rhel-che-devfile-registry`

cico_build.sh
```diff
@@ -69,17 +71,31 @@ function deploy() {
...
-  # Let's deploy
+  # Let's push 'rhel' images to 'quay.io/openshiftio'
   docker build -t ${IMAGE} -f ${DOCKERFILE} .
 
   TAG=$(echo "$GIT_COMMIT" | cut -c1-"${DEVSHIFT_TAG_LEN}")
 
   tag_push "${REGISTRY}/openshiftio/$IMAGE:$TAG"
   tag_push "${REGISTRY}/openshiftio/$IMAGE:latest"
-  echo 'CICO: Image pushed, ready to update deployed app'
+  echo "CICO: Images pushed to 'quay.io/openshiftio', ready to update deployed app"
```
l0rd: A few suggestions:
- I don't think you should login/push to `quay.io/eclipse-che` if the `"$TARGET" == "rhel" `
- You don't need to the `docker build` twice
- I would avoid tagging the images as `latest` because that should be done when we do the release of Che every 3 weeks (especially for eclipse-che org)

l0rd: And do we really need to push centos images to both openshiftio and eclipse-che?

ibuziuk: > I don't think you should login/push to quay.io/eclipse-che if the "$TARGET" == "rhel"
> You don't need to the docker build twice

Maybe I'm missing smth but in the issue description there is - `Update che-devfile-registry-master CentOS CI job to push rhel images to quay.io/openshiftio and centos images to quay.io/eclipse-che` and my understanding is that master CI job always build 'rhel for now, right?. So, I do not really understand how we can avoid building and pushing images twice if we want CentOS images beeing build as part of the same job

ibuziuk: it is actually related to my question in my first question in the PR - https://github.com/eclipse/che-devfile-registry/pull/31#issuecomment-510394593

l0rd: If what you are saying was true ($TARGET is always "rhel") the script would build `rhel` images **only** (the script does only one docker build). But we know that this is not the case because we see that centos images are pushed to the registry right? 

So I think that this script is called twice: once with not specific $TARGET and once with TARGET=rhel. And we should double check with SD team but I suspect that we (you and I) do not have the privileges to see the rhel images even if they are pushed in that org.

l0rd: Another thing I would add in this PR is an update to the README file to use the `eclipse-che` org images instead of the openshiftio.

ibuziuk: > So I think that this script is called twice: once with not specific $TARGET and once with TARGET=rhel

@l0rd looks like this is the case (script is called twice with & without the target), because:

- for now there is only one job for devfile registry - https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml#L4470
- centos images are pushed to https://quay.io/repository/openshiftio/che-devfile-registry?tab=tags
- on dsaas devfile-registry is build against `openshiftio/rhel-che-devfile-registry` images (and yes, we do not have access to https://quay.io/repository/openshiftio/rhel-che-devfile-registry)

ibuziuk:
@l0rd PR has been updated: 
- during the build rhel images would be pushed to `quay.io/openshiftio/rhel-che-devfile-registry`
- centos images will be pushed to `quay.io/eclipse-che/rhel-che-devfile-registry`
- https://quay.io/repository/openshiftio/che-devfile-registry should be deprecated

I plan to add yet another commit with docs & build instructions but before want to clarify your point around `latest` tag.
Basically, all the instructions in Readme are currently depended on this tag: e.g.:

- https://github.com/eclipse/che-devfile-registry/blob/master/build.sh#L11
- https://github.com/eclipse/che-devfile-registry/blob/master/deploy/openshift/che-devfile-registry.yaml#L89

if brand-new `quay.io/eclipse-che/rhel-che-devfile-registry` will not contain the latest tag, which one should be used ?

benoitf:
> centos images will be pushed to quay.io/eclipse-che/rhel-che-devfile-registry

why name is rhel-something if it's centos ?

ibuziuk:
@benoitf sorry, typo - updated

benoitf:
just a small note, on dockerhub we had `eclipse/che-something` as org was eclipse.
but on quay.io now with `eclipse-che` org, do we still `che` in image name ?

`eclipse-che/che-devfile-registry` --> `eclipse-che/devfile-registry` ?

ibuziuk:
@benoitf `che-devfile-registry` is the repo name so, I personally would expect to have the very same name for docker image, but I do understand your concern

benoitf:
@ibuziuk then it would mean quay org should the same than github org or that we move to eclipse-che org on github as well (it's possible to have its own 'eclipse' org so it could be a good idea as well so we cleary see only repositories linked to che)

ibuziuk:
@benoitf agree, https://quay.io/organization/eclipse would be more logical
@l0rd wdyt ?

l0rd:
I think eclipse organisation makes more sense as well. But didn't knew if we had the time to move to that org. 

For the change to `latest` I am proposing to:

1. stop tagging every build as `latest` (without actually changing the README etc...) => in this PR
2. start tagging nightlies with the new cron job and update the README etc...to use nigthlies
3. start tagging releases every 3 weeks as latest

But if you want to stay on the safe side (meaning that we want to make sure that the README will always point to the latest build) you should swap 1. and 2. and do that in separate PRs.

l0rd:
Anyway LGTM for the changes you have made so far

cico_build.sh
rhopp: Yep.. The script is run twice. Once here:
https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml#L891
And for the second time here (where `extra_target=rhel`):
https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml#L902

ibuziuk:
> I think eclipse organisation makes more sense as well. But didn't knew if we had the time to move to that org.

Indeed looks like we do not have time for moving from one org to anoter and setup ci bots for those

>stop tagging every build as latest (without actually changing the README etc...) => in this PR

Done

@l0rd sounds good, so no docs update in this PR. Could you please approve if you are ok to merge it in the current state.

P.S.

> So I think that this script is called twice: once with not specific $TARGET and once with TARGET=rhe

checked with SD and indeed `ci_cmd` command is executed twice (for `rhel` and `centos` builds) - https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml#L880-L909

ibuziuk:
https://quay.io/repository/eclipse-che/che-devfile-registry?tab=tags \o/
prod-preview has been also updated successfully \o/

PR #32, opened 2019-07-12
Add ASP.NET devfile

tolusha:
Signed-off-by: Anatoliy Bazko <abazko@redhat.com>

### What does this PR do?
Adds stack for developing ASP.NET Core web application
https://github.com/gothinkster/aspnetcore-realworld-example-app

### DIFF
--- a/devfiles/dotnet-asp.net/devfile.yaml
+++ b/devfiles/dotnet-asp.net/devfile.yaml
@@ -20,7 +20,7 @@ components:
   -
     type: dockerimage
     alias: dotnet
-    image: mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
+    image: quay.io/eclipse/che-dotnet-2.2:nightly
     command: ['sleep']
     args: ['infinity']
     env:
--- a/devfiles/dotnet-asp.net/devfile.yaml
+++ b/devfiles/dotnet-asp.net/devfile.yaml
@@ -7,7 +7,7 @@ projects:
     name: aspnetcore-example
     source:
       type: git
-      location: "https://github.com/gothinkster/aspnetcore-realworld-example-app"
+      location: "https://github.com/che-samples/aspnetcore-realworld-example-app"
 components:
   -
     type: chePlugin
--- a/devfiles/dotnet-asp.net/devfile.yaml
+++ b/devfiles/dotnet-asp.net/devfile.yaml
@@ -1,0 +1,67 @@
+---
+apiVersion: 1.0.0
+metadata:
+  name: asp.net
+projects:
+  -
+    name: aspnetcore-example
+    source:
+      type: git
+      location: "https://github.com/gothinkster/aspnetcore-realworld-example-app"
+components:
+  -
+    type: chePlugin
+    id: redhat-developer/che-omnisharp-plugin/latest
+    memoryLimit: 1024Mi
+  -
+    type: chePlugin
+    id: redhat-developer/netcoredbg-theia-plugin/latest
+    memoryLimit: 512Mi
+  -
+    type: dockerimage
+    alias: dotnet
+    image: mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
+    command: ['sleep']
+    args: ['infinity']
+    env:
+      - name: HOME
+        value: /home/user
+      - name: PS1
+        value: $(echo ${0})\\$
+    memoryLimit: 512Mi
+    endpoints:
+      - name: '5000'
+        port: 5000
+    mountSources: true
+    volumes:
+      - name: dotnet
+        containerPath: "/home/user"
+commands:
+  -
+    name: install Cake
+    actions:
+      - type: exec
+        component: dotnet
+        command: "dotnet tool install -g Cake.Tool"
+        workdir: ${CHE_PROJECTS_ROOT}/aspnetcore-example
+  -
+    name: build
+    actions:
+      - type: exec
+        component: dotnet
+        command: "$HOME/.dotnet/tools/dotnet-cake --runtime=linux-x64"
+        workdir: ${CHE_PROJECTS_ROOT}/aspnetcore-example
+  -
+    name: re-build
+    actions:
+      - type: exec
+        component: dotnet
+        command: "$HOME/.dotnet/tools/dotnet-cake --target=Rebuild --runtime=linux-x64"
+        workdir: ${CHE_PROJECTS_ROOT}/aspnetcore-example
+  -
+    name: run server
+    actions:
+      - type: exec
+        component: dotnet
+        command: "dotnet run --project src/Conduit/Conduit.csproj"
+        workdir: ${CHE_PROJECTS_ROOT}/aspnetcore-example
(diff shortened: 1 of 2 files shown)


l0rd:
cc @ibuziuk @amisevsk a new devfile to handle...

sparkoo:
I'm getting lot of errors, no code navigation... I'm using che-theia:next

asp_err

tolusha:
@sparkoo @svor 
Firstly try to install all dependencies.
So, follow the provided commands:
1. Install Cake
2. Build the project

sparkoo:
trying that, but editor is still unhappy :(

svor:
@sparkoo It can be because of https://github.com/eclipse/che/issues/13739
try to refresh the browser tab, after that omnisharp ls should pick up this project.

skabashnyuk:
404 is that expected?
Знімок екрана  о 13 54 05

svor:
@skabashnyuk preview URL should contain `/swagger` at the end. 
@tolusha need to find a way how to configure `previewUrl` in devfile. You can see how it was implemented in the stack: https://github.com/eclipse/che/blob/2b0149a01c67738faa590214057b60214454c854/ide/che-core-ide-templates/src/main/resources/samples.json#L1094

sparkoo:
devfile and it's tasks works for me. 
One issue with URL as @skabashnyuk mentioned.
Second issue with editor as @svor mentioned the issue https://github.com/eclipse/che/issues/13739. With this bug, user experience is terrible. The editor basically can't find any dependency and red-underline every reference, even to core .net api. After refresh, it loads properly and works. IMHO we can't have this in base samples. These must work out of the box, without any hurdles.

tolusha:
@skabashnyuk 
It is expected because nothing is configured at the root path. As @svor mentioned we need to open a `swagger` page (see the screenshot below). But I don't see the way how to configure the `previewUrl` for commands [1]. Any ideas?

[1] https://redhat-developer.github.io/devfile/

@sparkoo 
User is supposed to read the `readme.md` file before working on the project. The swagger URL is mentioned there. As a workaround it is possible to move the project to the `che-samples` and update the project by configuring some service at the root path and adding a link to the swagger page.

Screenshot from 2019-07-16 02-18-11

sparkoo:
@tolusha how swagger helps with the issue?

tolusha:
@sparkoo 
I meant there is no issue with 404. It is expected, because no services are configured on the root path.

sparkoo:
@tolusha ah, ok. I thought you were speaking about second issue with plugin. Make sense now :)
I guess we don't want to fork the project and I don't know how to set something like `previewUrl` either. I guess this sample isn't the only one with this issue. We should probably go through all of them and check whether we're satisfied with user experience we're delivering.

tolusha:
@sparkoo 
So, what is our plan toward this PR? postponed?

l0rd:
For the `previewUrl` problem I guess we do not have other option than forking the real-world example as a `che-sample` and make the root URL redirect to the application URL.

I don't know if there are other problems.

sparkoo:
@l0rd see my comment https://github.com/eclipse/che-devfile-registry/pull/32#issuecomment-511318510 There is issue with Omnisharp plugin https://github.com/eclipse/che/issues/13739

l0rd:
Ok but https://github.com/eclipse/che/issues/13739 is not related to this Devfile. That's about the support of any .NET project and is similar to https://github.com/eclipse/che/issues/13427 (that's about Java support). It can be related to the size of the project (how long it takes to clone it): if the cloning takes tool long a refresh will be needed. 

So @sparkoo you are suggesting not to include this sample until https://github.com/eclipse/che/issues/13427 is fixed right? I think it makes sense. Maybe in the meantime we can find out a smaller sample that would work.

sparkoo:
@l0rd yes, that's what I'm suggesting. Also C#.NET sample is affected by that issue, and I don't think we can go much simpler than that https://github.com/che-samples/dotnet-web-simple

l0rd:
@sparkoo ok I agree with you

amisevsk:
A number of existing devfiles also have the issue of the preview URL not opening to a path handled by the server (e.g. `nodejs-mongo`'s preview url just gives `{"errors":{"message":"Not Found","error":{}}}`, `PHP+MySQL` serves a directory structure instead of the app).

Is validity of endpoints/preview URLs something I should add to our checklist for testing devfiles?

sparkoo:
@amisevsk there is such a checklist? I've created an issue to retest and unify all devfiles once all are merged (https://github.com/eclipse/che/issues/13863), are you going to do that?

amisevsk:
@sparkoo I did a relatively quick run through in testing #38, available [here](https://gist.github.com/amisevsk/86ec3540da1b9b5ccd98aa23d4e9fafb), and @ScrewTSW is doing similar for che.openshift.io:  https://github.com/redhat-developer/che-functional-tests/issues/543#issuecomment-512795837

I personally have been glossing over preview URL / endpoint issues thus far, though. There are a number of potential improvements (e.g. some PHP devfiles do not expose the right port, the java stack exposes 8080 but runs `java-console-simple`, etc). My general bar has thus far been "I was able to make this work".

sparkoo:
@amisevsk ok, I didn't know about this effort. My opinion is, that these are our samples and they must work perfectly. URL that is loaded after project start is public facing and very visible. It is highly probably that user will hit it, so I would consider it as a blocker issue.
Imagine you're trying Che for the first time, trying sample with your language of choice and you hit the issue. What is the chance you'll try your project and consider Che as your main IDE? When even samples from authors does not work 100% ...
That's only my opinion, though. It might not be that important.

l0rd:
@sparkoo perfectly agree with your idea. To address this now we have only one approach: modify/fork the sample to make them work with the current endpoint URL (with no subpath). I have created https://github.com/eclipse/che/issues/13945 to support that better in the next sprints.

@amisevsk yes you should track the problems with the URL as well. And if we could automate the generation of the report and add it as a PR check it would be awesome :upside_down_face:

devfiles/dotnet-asp.net/devfile.yaml
```diff
@@ -0,0 +1,67 @@
...
+    id: redhat-developer/che-omnisharp-plugin/latest
+    memoryLimit: 1024Mi
+  -
+    type: chePlugin
+    id: redhat-developer/netcoredbg-theia-plugin/latest
+    memoryLimit: 512Mi
+  -
+    type: dockerimage
+    alias: dotnet
+    image: mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
```
amisevsk: Before merging this PR, please update this to use one of the patched images (`quay.io/eclipse/che-dotnet-2.2:nightly`)

ibuziuk:
@l0rd since there are problems with URL / subpath could we avoid merging this PR for GA and postpone it till 7.1.0 ?

l0rd:
@ibuziuk yes that makes sense. But let's merge this if we are able to fix the URL problems in the samples before the GA.

tolusha:
Image for dev container is updated.
Redirection to a swagger page is added.

tolusha:
@ibuziuk 
Is it possible to merge now?

svor:
Redirection to `/swagger` resources works fine on my side.

ibuziuk:
@tolusha I would rather postpone it to `7.1.0` similar to C / C++ one, wdyt ?

tolusha:
@ibuziuk 
Ok make sense.

PR #35, opened 2019-07-12
che #13813: Adding script for the nightly ci build

ibuziuk:
### What does this PR do?
Adding a separate script for the nightly ci build

https://github.com/eclipse/che/issues/13813 
the second part of the issue - `Update che-devfile-registry-periodic CentOS CI job to build the devfile registry image (centos only) and tag it as nightly`

### DIFF
--- a/cico_build_nightly.sh
+++ b/cico_build_nightly.sh
@@ -1,0 +1,79 @@
+#!/bin/bash
+#
+# Copyright (c) 2012-2018 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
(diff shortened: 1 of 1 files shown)


l0rd:
The script looks identical to the ci one (except for the TAG). Isn't it a way to provide the TAG as a parameter?

ibuziuk:
@l0rd it is possible to do but this will overcomplicate the `cico_build.sh` in future since I plan to extend the `cico_build_nightly.sh` building also  https://github.com/eclipse/che-devfile-registry/tree/master/arbitrary-users-patch

In general, supporting cico scripts is not straight so I personally prefer to have distinct scripts for distinct jobs. IMO it will much simplify the support burden since for example for nightly there will be a separate job template in https://github.com/openshiftio/openshiftio-cico-jobs/blob/master/devtools-ci-index.yaml

ibuziuk:
Related cico job PR - https://github.com/openshiftio/openshiftio-cico-jobs/pull/1037

PR #37, opened 2019-07-12
che #13813: Updating docs and scripts to point to the brand-new 'quay.io/eclipse/che-devfile-registry' which is the default one from now on

ibuziuk:
### What does this PR do?
Updating docs and scripts to point to the brand-new 'quay.io/eclipse/che-devfile-registry' which is the default one from now on

Related to https://github.com/eclipse/che/issues/13813


ibuziuk:
Closing since this one is probably outdated for now

PR #42, opened 2019-07-20
Update dockerfile patch and make new java devfiles used patched images

amisevsk:
### What does this PR do?
- Update the arbitrary users patch to make sure $HOME is always writable,
fixing issues with some images.

- Update `java-spring`, `java-mysql`, and `java-vertx` to use patched images
    - Update the commands in these images since maven wrapper is incompatible with community maven image
    - The base image used for these stacks is **changed** in this PR: we use `maven:3.6.1-jdk-8` instead of `registry.centos.org/che-stacks/centos-jdk8` to avoid building another patched image.

One change this makes for the spring images is changing the command to **not** use the maven wrapper. This wrapper is intended to make it easier to use maven (https://github.com/takari/maven-wrapper/) but appears to not be necessary for the image we are using. Additionally the wrapper is incompatible with the maven community image on OpenShift, since the community image defines `$MAVEN_CONFIG=/root/.m2`, and the wrapper appears to reuse this env var for (undocumented, afaict) reasons: https://github.com/takari/maven-wrapper/blob/master/mvnw#L301

Part of https://github.com/eclipse/che/issues/13454, https://github.com/redhat-developer/rh-che/issues/1455, some issues raised in #38 (gradle home not writable)

### Tested images:

- `java-vertx`, `java-spring`, `java-mysql`: Everything works except:
   1. Issue https://github.com/eclipse/che/issues/13938 is present (in patched and current master images -- i.e. this PR does not fix this issue)
   2. As in some other stacks, a `.bashrc` already exists so these stacks get an ugly prompt (including workspace ID)

- `java-gradle`: This PR fixes the `/home/gradle` directory not being writable

- `php-simple`: This PR fixes the `/home/user` directory not being writable.

### DIFF
--- a/arbitrary-users-patch/Dockerfile
+++ b/arbitrary-users-patch/Dockerfile
@@ -2,7 +2,9 @@ ARG FROM_IMAGE
 FROM ${FROM_IMAGE}
 USER 0
 # Set permissions on /etc/passwd and /home to allow arbitrary users to write
-RUN chmod g=u /etc/passwd /home
+RUN mkdir -p /home/user
+RUN chgrp -R 0 /home
+RUN chmod -R g=u /etc/passwd /home
 COPY [--chown=0:0] entrypoint.sh /
 RUN chmod +x /entrypoint.sh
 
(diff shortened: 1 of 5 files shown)


amisevsk:
@sleshchenko Do you foresee any subtle issues resulting from us no longer using the maven wrapper to build the Spring projects? I've tested with default `mvn clean install` and it seems to work without issue.

PR #47, opened 2019-07-24
che-13953: Adding a simple mechanism to automatically release devfile registry using the existing 'build-master' CI

ibuziuk:
### What does this PR do?
Adding a simple mechanism to automatically release devfile registry using the existing 'build-master' CI:

existing 'build-master' CI https://ci.centos.org/job/devtools-che-devfile-registry-build-master/ will do a release if needed based on the VERSION file. If the version is not a SNAPSHOT one CI will also tag & push 'che-devfile-registry' based on the version defined in the VERSION file

https://github.com/eclipse/che/issues/13953


ibuziuk:
@l0rd ok, sounds good - I will stick to the initial plan then. I just thought that the approach with releasing from the `master` is much more straight-forward for dev-file-registry.

ibuziuk:
Closed in favour of - https://github.com/eclipse/che-devfile-registry/pull/49

PR #49, opened 2019-07-24
che-13953: Adding cico release script for CI

ibuziuk:
### What does this PR do?
Adding cico release script for CI that will tag and push  1) devfile registry image 2) all `arbitary-user-patch` images based on the version defined in the VERSION file

NOTE: even though the script would be located in the master branch initially CI job will be triggered from the dedicated `release` branch:

the idea is to create a `release` branch from `master` (e.g. `7.1.0`) we do one or 2 commits (set version, set versions of the patched images) and when we are happy we merge it to a branch named `release` that corresponds to the  special `*-release job` in the CI.
 
 
eclipse/che#13953


skabashnyuk:
CC @vparfonov

cico_release.sh
```diff
@@ -0,0 +1,85 @@
...
+  TAG=$(head -n 1 VERSION)
+
+  if [ -n "${QUAY_ECLIPSE_CHE_USERNAME}" ] && [ -n "${QUAY_ECLIPSE_CHE_PASSWORD}" ]; then
+    docker login -u "${QUAY_ECLIPSE_CHE_USERNAME}" -p "${QUAY_ECLIPSE_CHE_PASSWORD}" "${REGISTRY}"
+  else
+    echo "Could not login, missing credentials for pushing to the '${ORGANIZATION}' organization"
+  fi
+
+  "${SCRIPT_DIR}"/arbitrary-users-patch/build_images.sh --push
+  echo "CICO: pushed '${TAG}' version of the arbitrary-user patched base images"
```
amisevsk: This will tag arbitrary UID images but the devfiles will still point to `nightly`; we would need some way of automatically updating tags used in devfiles as well.

amisevsk: Can be done manually pretty easily, but would be another manual step to remember.

ibuziuk: yes, but for now it is not required (we discussed it with @l0rd)

PR #61, opened 2019-07-25
Fixing Java MySQL sample devfile to be able to be run on OSIO

ScrewTSW:
### What does this PR do?
This PR fixes the Java MySQL devfile memory quotas in order to be able to run the devfile on OSIO

Issue: https://github.com/redhat-developer/che-functional-tests/issues/556

### DIFF
--- a/devfiles/java-mysql/devfile.yaml
+++ b/devfiles/java-mysql/devfile.yaml
@@ -12,6 +12,7 @@ components:
   -
     type: chePlugin
     id: redhat/java/latest
+    memoryLimit: 1280MiB
   -
     type: dockerimage
     alias: tools
@@ -45,7 +46,7 @@ components:
         value: petclinic
       - name: PS1
         value: $(echo ${0})\\$
-    memoryLimit: 256Mi
+    memoryLimit: 300Mi
     endpoints:
       - name: 'db'
         port: 3306
--- a/devfiles/java-mysql/meta.yaml
+++ b/devfiles/java-mysql/meta.yaml
@@ -3,4 +3,4 @@ displayName: Java with Spring Boot and MySQL
 description: Java stack with OpenJDK 8, MySQL and Spring Boot Petclinic demo application
 tags: ["Java", "OpenJDK", "Maven", "Spring Boot", "MySQL"]
 icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
-globalMemoryLimit: 2930Mi
+globalMemoryLimit: 3Gi


devfiles/java-mysql/devfile.yaml
```diff
@@ -24,7 +25,13 @@ components:
           -Duser.home=/home/user"
       - name: MAVEN_OPTS
         value: $(JAVA_OPTS)
+<<<<<<< Updated upstream
```
skabashnyuk: is this line expected to be here?

devfiles/java-mysql/devfile.yaml
```diff
@@ -24,7 +25,13 @@ components:
...
     memoryLimit: 700Mi
+=======
```
skabashnyuk: is this line expected to be here?

devfiles/java-mysql/devfile.yaml
```diff
@@ -24,7 +25,13 @@ components:
...
+      - name: PS1
+        value: $(echo ${0})\\$
+    memoryLimit: 768Mi
+>>>>>>> Stashed changes
```
skabashnyuk: is this line expected to be here?

devfiles/java-mysql/devfile.yaml
ScrewTSW: that's weird. looks as if VisualStudioCode didn't rewrite the file
I'll fix it

devfiles/java-mysql/devfile.yaml
ScrewTSW: fixed

devfiles/java-mysql/devfile.yaml
ScrewTSW: fixed

skabashnyuk:
CC @amisevsk

devfiles/java-mysql/devfile.yaml
```diff
@@ -24,7 +25,9 @@ components:
           -Duser.home=/home/user"
       - name: MAVEN_OPTS
         value: $(JAVA_OPTS)
-    memoryLimit: 700Mi
+      - name: PS1
+        value: $(echo ${0})\\$
```
amisevsk: `PS1` should not be included, it is set in the entrypoint on the patched image.

amisevsk: [ref](https://github.com/eclipse/che-devfile-registry/blob/92624598ba6cb07007bae1f0281262fa93e5a34b/arbitrary-users-patch/entrypoint.sh#L10)

amisevsk:
Looking at my quota usage with this devfile, I see `2.77Gi` used out of `3Gi` available; is there a reason we're not using the remaining `~230Mi` quota?

devfiles/java-mysql/devfile.yaml
ScrewTSW: will remove :)

ScrewTSW:
> Looking at my quota usage with this devfile, I see `2.77Gi` used out of `3Gi` available; is there a reason we're not using the remaining `~230Mi` quota?

I can try to fiddle with it and submit a fix to the PR tomorrow.

amisevsk:
Regardless I've tested and the lowered limit on java LS does not seem to interfere for the sample project. +1 to merge.

amisevsk:
Looking at this more closely (output of `top` on the vscode-java container) I can see the language server being killed occasionally and restarted by the plugin. This is accompanied by logs like
```
[Info  - 2:59:18 PM] Connection to server got closed. Server will restart.
[Info  - 2:59:56 PM] Connection to server got closed. Server will restart.
[Error - 2:59:56 PM] Request textDocument/codeAction failed.
Error: Connection got disposed.
[...]
```

I also at one point ended up with two java language servers running simultaneously (though I don't know if that's an issue in `master`)

ScrewTSW:
so upon investigation there are two constants consuming quota limits:
* JWT Proxy consumes 150MiB
* Machine exec pod consumes 150MiB
* Theia consumes **512MB** = 488.281MiB

and the plugins defined in the devfile:
* mysql : 256MiB (+40MiB new value **296MiB**)
* Java LSP : 1024MiB
* tools (maven) : 768MiB (+192MiB new value **960MiB**)

That leaves me 235.71875MiB to redistribute. After redistribution this leaves me exactly 3899392 bytes to spare or 3.71875MiB .

PR #62, opened 2019-07-25
Fixed java-web-spring quotas to be able to run on OSIO

ScrewTSW:
### What does this PR do?
This PR fixes the Java Springboot devfile memory quotas in order to be able to run the image on OSIO

Issue: https://github.com/redhat-developer/che-functional-tests/issues/556

### DIFF
--- a/devfiles/java-web-spring/devfile.yaml
+++ b/devfiles/java-web-spring/devfile.yaml
@@ -12,6 +12,7 @@ components:
   -
     type: chePlugin
     id: redhat/java/latest
+    memoryLimit: 1512Mi
   -
     type: dockerimage
     alias: tools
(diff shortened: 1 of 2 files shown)


ScrewTSW:
so upon investigation there are three constants consuming quota limits:
* JWT Proxy consumes 150MiB
* Machine exec pod consumes 150MiB
* Theia consumes **512MB** = 488.281MiB

and the plugins defined in the devfile:
* Java LSP : **(value unset)** ->  changed to **1512MiB** - Should be enough to avoid LSP OOM restart
* tools (maven) : 1024MiB (pointlessly large,-256MiB new value **768MiB**)

That leaves me 1259.71875MiB to work with. After redistribution this leaves me exactly 3899392 bytes to spare or 3.71875MiB .

Java LSP peaked at 1258MiB
tools (maven) peaked at 385MiB even after rigorous usage

devfiles/java-web-spring/devfile.yaml
```diff
@@ -12,6 +12,7 @@ components:
   -
     type: chePlugin
     id: redhat/java/latest
+    memoryLimit: 1512Mi
```
amisevsk: This *could* potentially be dropped, as the default used for `redhat/java/latest` is `1500Mi`. However this would mean any increases to the plugin-registry default would break this devfile again.

PR #63, opened 2019-07-25
Always build rhel images and patched images

l0rd:
Rhel images should be build when we do releases because it's better to
use release vs nightlies in prod.

Building patched images in ci build so that as soon as a PR is merged we
can use the new patched images as well (not only the new registry
image).

I have also done some refactoring:
- removed duplicate code
- renamed the script files consistently

This PR should be merged only when the cico jobs have been adapted: [PR](https://github.com/openshiftio/openshiftio-cico-jobs/pull/1049)


cico_functions.sh
```diff
@@ -46,17 +46,36 @@ function install_deps() {
...
+
+function set_ci_tag() {
+  # Let's obtain the tag based on the 
+  # version defined in the 'VERSION' file
+  TAG=$(echo "$GIT_COMMIT" | cut -c1-"${DEVSHIFT_TAG_LEN}")
+  export TAG
+}
+
+function set_nightly_tag() {
+  # Let's obtain the tag based on the
```
ibuziuk: @l0rd the comment looks incorrect

cico_functions.sh
```diff
@@ -46,17 +46,36 @@ function install_deps() {
...
 
+function set_release_tag() {
+  # Let's obtain the tag based on the 
+  # version defined in the 'VERSION' file
+  TAG=$(head -n 1 VERSION)
+  export TAG
+}
+
+function set_ci_tag() {
+  # Let's obtain the tag based on the
```
ibuziuk: @l0rd the comment looks incorrect

cico_functions.sh
l0rd: fixed

cico_functions.sh
l0rd: fixed

ibuziuk:
@l0rd PR with cico jobs has been merged so, we need to update scripts names now - https://github.com/eclipse/che-devfile-registry/pull/64

l0rd:
@ibuziuk please approve or I cannot merge otherwise

PR #66, opened 2019-07-30
Add debug configurations to java server devfiles

amisevsk:
### What does this PR do?
Sets up java-spring and java-vertx devfiles to support debugging.

- Adds a default debug config to each devfile so that clicking "Start debugging" works out of the box
- Modifies "run app" commands to start with debugging enabled

With these changes, the flow of "run webapp" -> "Start debugging" works without additional configuration.

https://github.com/redhat-developer/rh-che/issues/1434 (although this is not an osio-specific issue)

### DIFF
--- a/devfiles/apache-camel-springboot/devfile.yaml
+++ b/devfiles/apache-camel-springboot/devfile.yaml
@@ -63,3 +63,26 @@ commands:
       component: maven
       command: mvn spring-boot:run -DskipTests
       workdir: ${CHE_PROJECTS_ROOT}/fuse-rest-http-booster
+  -
+    name: run the services (debugging enabled)
+    actions:
+      - type: exec
+        component: maven
+        command: mvn spring-boot:run -DskipTests -Drun.jvmArguments="-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005"
+        workdir: ${CHE_PROJECTS_ROOT}/fuse-rest-http-booster
+  -
+    name: Debug remote java application
+    actions:
+      - type: vscode-launch
+        referenceContent: |
+          {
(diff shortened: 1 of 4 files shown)


amisevsk:
Added the same fixes for the apache-camel devfile.

amisevsk:
@apupier I agree that a one button "start debugging" would be ideal, but wouldn't using the debug codelens require opening the correct file? I only get the codelens on some files. As far as I know there isn't a direct way to combine a launch task with launching a debug configuration, and using just theia launches has given me nothing but trouble (especially since theia and the workspace run in separate containers).

In terms of naming, I think the names provided are fairly clear (start app to start the server, debug remote (attach) to attach your debugger). I can work on improving these if you have any suggestions.

For port 5005, no there's no fix at the moment. I've left a comment on issue https://github.com/eclipse/che/issues/13990, which is somewhat related and has been prioritized for `7.x`.

apupier:
> but wouldn't using the debug codelens require opening the correct file?

yes

> I think the names provided are fairly clear (start app to start the server, debug remote (attach) to attach your debugger). I can work on improving these if you have any suggestions.

"run the services (debugging enabled)" --> for me I really understood it as the debugger is already working. This is something which seems so natural with todays IDE that even if it is not "debugging activated" and only "debugging enabled", I understood it this way.

"Debug (Attach) - Remote", for an end-user of the IDE, there is no reason to have a "remote" connection. From user point of view, it is "local to the Che workspace". This is really an internal detail implementation. I don't think that end-users will understand that it will connect to the process launched with "run the services (debugging enabled)" just through the UI.

I propose `run the services (ready to attach the remote debugger)` or `run the services (ready to "Debug (Attach) - Remote")`

PR #67, opened 2019-07-30
Make devfile endpoints match their sample apps

amisevsk:
### What does this PR do?
Update devfiles so that exposed ports match those used by the sample app:
- golang: add port `8080`
- java-gradle: remove unused port (simple console app)
- java-maven: remove unused port (simple console app)
- php-laravel: update port to `8000` (instead of `8080`)
- php-symfony: update port to `8000` (instead of `8080`)
- php-web-simple: Add endpoint for port `8080`

https://github.com/redhat-developer/rh-che/issues/1434 (though not osio-specific)

### DIFF
--- a/devfiles/go/devfile.yaml
+++ b/devfiles/go/devfile.yaml
@@ -28,6 +28,9 @@ components:
       # replicate the GOCACHE from the plugin, even though the cache is not shared
       # between the two
       value: /tmp/.cache
+  endpoints:
+    - name: '8080/tcp'
+      port: 8080
   memoryLimit: 512Mi
   mountSources: true
 commands:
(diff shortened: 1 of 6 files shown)


l0rd:
@ibuziuk @amisevsk how is this related to https://github.com/eclipse/che/issues/13964?

amisevsk:
@l0rd The main relation is in the php and golang devfiles; those do not expose the port used by their sample application by default. While Theia has the autoexpose feature, it pops up as a prompt at the top of the window, and if you miss or close this prompt you cannot access your sample app easily. Additionally, there are some minor UX issues with the autoexpose that I think we should avoid for sample applications:
- You have to expose the port via the prompt every time the server starts (so restarting the server is a little more tedious)
- It takes a second for the port to expose; on opening the preview after autoexposing, it takes ~5 seconds before the route updates. In the mean time it shows the OpenShift "Application not available" window.

With these changes you're always able to open the preview and view the running app.

For https://github.com/eclipse/che/issues/13964, there is still a problem with three the three devfiles, where the default endpoint is not applicable for the application; I'm looking at those next.

PR #68, opened 2019-07-31
Add debug config and stop command for go devfile

amisevsk:
### What does this PR do?
Improves the go devfile:
- Adds a default debug config that makes it easier to debug the current file.
- Adds "Stop outyet" task, since otherwise stopping the server either requires executing `kill` in a terminal or sending `ctrl+C` to the `run outyet` task's tab (which is entirely empty)
  - Also stopping the running task is necessary to start debugging

https://github.com/redhat-developer/rh-che/issues/1434 (Debugging working is in the acceptance criteria)

### DIFF
--- a/devfiles/go/devfile.yaml
+++ b/devfiles/go/devfile.yaml
@@ -5,7 +5,7 @@ metadata:
 projects:
 -
   name: example
-  source: 
+  source:
     type: git
     location: https://github.com/golang/example.git
   clonePath: src/github.com/golang/example/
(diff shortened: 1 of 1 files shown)


amisevsk:
Note also that in the go devfile, debugging only works when Theia launches the relevant app (it has to run in the go plugin sidecar since delve is not available in the main container.

This may cause issues with other projects, but works for `outyet` in my testing.

l0rd:
@amisevsk can you clarify your comment above: "only works when Theia launches the relevant app". What do you mean? When does it fail? And how is it related to https://github.com/eclipse/che/issues/13816?

amisevsk:
@l0rd:

> clarify your comment above: "only works when Theia launches the relevant app". What do you mean? When does it fail? And how is it related to eclipse/che#13816?

It's a similar problem to eclipse/che#13816: 
- The theia container cannot run the go app because it does not have dependencies; as a result, it's necessary to debug remotely (where the remote is another container)
- Debugging in the main workspace container is not possible, as `delve` is not available and there is no configuration. Fixing this would require updating our base container and figuring out what configuration is necessary.
- This means that debugging *has* to happen in the go plugin container.
    - This is more prone to failure because the sample app has to run alongside the go plugin, limited by the plugins default memory limit (`512Mi`). It could be killed by the container or the entire container could reach an OOMKill
   - It also means that memory requirements are duplicated between the main container and the plugin sidecar; if your app requires `512Mi`, then both the main dev container and the plugin sidecar need to accomodate that
   - Finally, it means that you can only start debugging through Theia, since it's not easy to do things in the golang sidecar. As contrast, what I've tried to do for e.g. the java devfiles, is set it up so that you can run the application normally (in the main container) and automatically connect theia remote debugging to it.

Also related is https://github.com/eclipse/che/issues/14091, as it seems most of the default (non-remote) debug configs in Theia just won't work when running in Che. Setting debugging up can be counterintuitive and requires an understanding of Che infrastructure sometimes.

amisevsk:
@l0rd I realized that description could be unclear, so I drew a diagram of the three cases we have now:

Che-debugging

PR #70, opened 2019-08-01
Update meta.yamls to use Mi instead of Gi

amisevsk:
### What does this PR do?
Workaround for issue https://github.com/eclipse/che/issues/14096

Devfile `meta.yaml`s which specify `Gi` in their `globalMemoryLimit` display as `0Gi` in the stacks tab.

Screenshot from 2019-08-01 09-24-20

https://github.com/eclipse/che/issues/14096


ibuziuk:
@l0rd @skabashnyuk @vparfonov @rhopp this is 0 risk PR could we get your approval and merge it to master?

PR #71, opened 2019-08-02
Fix devfile for dotnet

svor:
Signed-off-by: Valeriy Svydenko <vsvydenk@redhat.com>

### What does this PR do?
Makes it possible to debug an application by adding several improvements:
-  The project was updated, now it is using netcoreapp2.2
-  Added a command to build the project. It is needed to be run before running debug
- Predefined debug configuration

https://github.com/eclipse/che/issues/14090

### Demo:
https://youtu.be/0xYFGMwPxPc

### DIFF
--- a/devfiles/dotnet/devfile.yaml
+++ b/devfiles/dotnet/devfile.yaml
@@ -8,13 +8,16 @@ projects:
     source:
       type: git
       location: "https://github.com/che-samples/dotnet-web-simple"
+      branch: netcoreapp2.2
 components:
   -
     type: chePlugin
+    alias: omnisharp
     id: redhat-developer/che-omnisharp-plugin/latest
     memoryLimit: 1024Mi
   -
     type: chePlugin
+    alias: netcoredbg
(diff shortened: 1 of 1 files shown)


slemeur:
Good work @svor !

devfiles/dotnet/devfile.yaml
```diff
@@ -44,3 +47,28 @@ commands:
         component: dotnet
         command: "dotnet run"
         workdir: ${CHE_PROJECTS_ROOT}/dotnet-web-simple
+  -
+    name: build
+    actions:
+    - type: exec
+      component: netcoredbg
```
amisevsk: Since in both cases it's building into `/projects` (mounted to all containers) and using the resulting `obj/Debug` folder, can this command be attached to the main ws container? As it is, it's hidden in the normally collapsed "plugins" tab in the "my workspace" sidebar.

svor: Makes sense.  I've added an improvement.

PR #76, opened 2019-08-09
use generateName instead of just name

sparkoo:
Signed-off-by: Michal Vala <mvala@redhat.com>

### What does this PR do?
updates devfiles to use `generateName` instead of just name.
Targeted to v7.1.0

https://github.com/eclipse/che/issues/13794
merge after this PR https://github.com/eclipse/che/pull/14157 !!!

### DIFF
--- a/devfiles/angular/devfile.yaml
+++ b/devfiles/angular/devfile.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: 1.0.0
 metadata:
-  name: angular
+  generateName: angular-
 projects:
   -
     name: angular-realworld-example-app
--- a/devfiles/apache-camel-springboot/devfile.yaml
+++ b/devfiles/apache-camel-springboot/devfile.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: 1.0.0
 metadata:
-  name: apache-camel-springboot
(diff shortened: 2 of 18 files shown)


sparkoo:
@l0rd yes. Current implementation in https://github.com/eclipse/che/pull/14157 is without programmed separator and that's how Kubernetes do it. We can do it with separator and I think it would be cleaner. What do you think @sleshchenko @skabashnyuk @benoitf ?

skabashnyuk:
I would stay to k8s api as close as possible.  For example. https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/ . That means if a user wants "-" he has to explicitly add it.

sleshchenko:
@sparkoo I'm OK with the current implementation where a user should include separator in `generateName` if he needs it (K8s API style).

benoitf:
I prefer k8s implementation (but only update doc saying that it's ```[generateName]YYYYY``` and not `[generateName]-YYYYY`)

BTW we could still use in examples generateName without a leading dash as well

as a user I mostly don't care to have for my workspace named `che1234` or `che-1234`or che_1234`

but also in k8s examples most of generateName have leading `-`

PR #77, opened 2019-08-09
Add command to configure DocumentRoot in php devfiles

amisevsk:
### What does this PR do?
Adds a command to simpler php devfiles to configure `DocumentRoot` and avoid showing an index page when the application is launched.

Without PR:
Screenshot from 2019-08-09 10-10-54

With PR (after configure task has been executed)
Screenshot from 2019-08-09 10-47-19

https://github.com/eclipse/che/issues/13964


ibuziuk:
@amisevsk could you please resolve conflicts?

PR #78, opened 2019-08-09
#14076 don't need yq; clean up shell scripts and Dockerfiles

nickboldt:
1. #14076 don't actually need yq since we're just looking for toplevel yaml fields; can use sed with '^fieldName:' to make the container more lightweight

2. clean up Dockerfiles and switch to './index.sh devfiles' (more consistent with './check_mandatory_fields.sh devfiles')

I know some people don't like grep|sed but this is way faster than using either yq implementation, and means it'll be easier to move this container to a ubi8-minimal base.

Change-Id: I76edb47e5c407e879467b9f02ccf52a50b35efaa

### DIFF
--- a/Dockerfile.rhel
+++ b/Dockerfile.rhel
@@ -12,23 +12,22 @@ RUN apk add --no-cache bash
 COPY .htaccess README.md *.sh /build/
 COPY /devfiles /build/devfiles
 WORKDIR /build/
-RUN ./check_mandatory_fields.sh devfiles
-RUN ./index.sh > /build/devfiles/index.json
+RUN ./check_mandatory_fields.sh devfiles && \
+    ./index.sh devfiles > /build/devfiles/index.json
 
 FROM quay.io/openshiftio/rhel-base-httpd:latest
 
-RUN sed -i -e "s,Listen 80,Listen 8080," /etc/httpd/conf/httpd.conf
-RUN sed -i -e "s,logs/error_log,/dev/stderr," /etc/httpd/conf/httpd.conf
-RUN sed -i -e "s,logs/access_log,/dev/stdout," /etc/httpd/conf/httpd.conf
-RUN sed -i -e "s,AllowOverride None,AllowOverride All," /etc/httpd/conf/httpd.conf
-
 EXPOSE 80
 
-# the htpasswd file may be periodically replaced during run
-RUN chmod a+rwX            /etc/httpd/conf
-RUN chmod a+rwX            /run/httpd
+RUN sed -e "s,Listen 80,Listen 8080," \
+    -e "s,logs/error_log,/dev/stderr," \
+    -e "s,logs/access_log,/dev/stdout," \
+    -e "s,AllowOverride None,AllowOverride All," \
+    -i /etc/httpd/conf/httpd.conf && \
+    # the htpasswd file may be periodically replaced during run
+    chmod a+rwX /etc/httpd/conf /run/httpd && \
+    mkdir -p /var/www/html/devfiles
 
-RUN mkdir /var/www/html/devfiles
 COPY --from=builder /build/ /var/www/html/
 USER 0
 RUN chmod -R g+rwX /var/www/html/devfiles
--- a/check_mandatory_fields.sh
+++ b/check_mandatory_fields.sh
@@ -10,44 +10,41 @@
 
 set -e
 
-FIELDS=('displayName' 'description' 'tags' 'icon' 'globalMemoryLimit')
+# Arguments:
+# 1 - folder in which to search for meta.yaml files
+if [[ ! $1 ]]; then echo "Error: must specify path to devfiles, eg., $0 /path/to/che-devfile-registry/devfiles"; exit 1; fi
+devfilesDir=$1; devfilesDir=${devfilesDir%/} # trim trailing slash if present
 
-
+metaInfoFields=('displayName' 'description' 'tags' 'icon' 'globalMemoryLimit')
 
 # check that field value, given in the parameter, is not null or empty
 function check_field() {
-  if [[ $1 == "null" || $1 = "" ]];then
+  if [[ $1 == "null" || $1 = "" ]]; then
     return 1;
   fi
   return 0
 }
 
-readarray -d '' arr < <(find "$1" -name 'meta.yaml' -print0)
-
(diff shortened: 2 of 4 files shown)


check_mandatory_fields.sh
```diff
@@ -10,44 +10,41 @@
...
 set -e
 
-FIELDS=('displayName' 'description' 'tags' 'icon' 'globalMemoryLimit')
-
+# Arguments:
+# 1 - folder in which to search for meta.yaml files
+if [[ ! $1 ]]; then echo "Error: must specify path to devfiles, eg., $0 /path/to/che-devfile-registry/devfiles"; exit 1; fi
+devfilesDir=$1; devfilesDir=${devfilesDir%/} # trim trailing slash if present
 
+metaInfoFields=('displayName' 'description' 'tags' 'icon' 'globalMemoryLimit')
```
nickboldt: Instead of FIELDS, use metaInfoFields (as in other shell script)

check_mandatory_fields.sh
```diff
@@ -10,44 +10,41 @@
...
     unset NULL_OR_EMPTY_FIELDS
-
-    for FIELD in "${FIELDS[@]}"
-    do
-      VALUE=$(yq r "$i" "$FIELD")
-
-      if ! check_field "${VALUE}";then
-        NULL_OR_EMPTY_FIELDS+="$FIELD "
+    for field in "${metaInfoFields[@]}"; do
+      value=$(grep "${field}:" "${i}" | sed -e "s#^${field}: ##" -e 's#"##g')
```
nickboldt: Don't need yq for something this simple. Just look for lines starting with (^) {field} and a colon, then strip out quotation marks.

check_mandatory_fields.sh
```diff
@@ -10,44 +10,41 @@
...
-
-for i in "${arr[@]}"
-do
-    id=$(yq r "$i" displayName | sed 's/^"\(.*\)"$/\1/')
-    full_id=${id}:${i}
-
+## loop through meta.yaml files
+for i in $(ls -1 "${devfilesDir}/"*"/meta.yaml" | sort); do
     echo "Checking devfile '${i}'"
+    id=$(grep "displayName:" "${i}" | sed -e "s#^displayName: ##" -e 's#"##g')
```
nickboldt: Don't need yq for something this simple. Just look for lines starting with (^) {field} and a colon, then strip out quotation marks.

index.sh
```diff
@@ -11,32 +11,32 @@
...
+function buildIndex() {
     FIRST_LINE=true
     echo "["
 
-    ## now loop through meta.yaml files
-    for i in "${arr[@]}"
-    do
-        if [ "$FIRST_LINE" = true ] ; then
+    ## loop through meta.yaml files
+    for i in $(ls -1 "${devfilesDir}/"*"/meta.yaml" | sort); do
```
nickboldt: find requires findutils RPM to be installed; `ls` does the same job with smaller container.

index.sh
```diff
@@ -47,4 +47,4 @@ function buildIndex() {
     echo "]"
 }
 
-buildIndex devfiles
+buildIndex
```
nickboldt: To be consistent with `check_mandatory_fields.sh` and other scripts (like those in plugin-registry/scripts/ folder) the path to be queried should be passed in as cmdline var from the Dockerfile invocation.

Dockerfile.rhel
```diff
@@ -12,23 +12,22 @@ RUN apk add --no-cache bash
...
-RUN sed -i -e "s,logs/error_log,/dev/stderr," /etc/httpd/conf/httpd.conf
-RUN sed -i -e "s,logs/access_log,/dev/stdout," /etc/httpd/conf/httpd.conf
-RUN sed -i -e "s,AllowOverride None,AllowOverride All," /etc/httpd/conf/httpd.conf
-
 EXPOSE 80
 
-# the htpasswd file may be periodically replaced during run
-RUN chmod a+rwX            /etc/httpd/conf
-RUN chmod a+rwX            /run/httpd
+RUN sed -e "s,Listen 80,Listen 8080," \
```
nickboldt: combine 6 steps into one operation.

Dockerfile.rhel
```diff
@@ -12,23 +12,22 @@ RUN apk add --no-cache bash
 COPY .htaccess README.md *.sh /build/
 COPY /devfiles /build/devfiles
 WORKDIR /build/
-RUN ./check_mandatory_fields.sh devfiles
-RUN ./index.sh > /build/devfiles/index.json
+RUN ./check_mandatory_fields.sh devfiles && \
+    ./index.sh devfiles > /build/devfiles/index.json
```
nickboldt: combine 2 steps into one

nickboldt:
Using sed:
* 0.385s for ./check_mandatory_fields.sh
* 0.246s for ./index.sh

Using yq (mikefarah version):
* 0.38s for ./check_mandatory_fields.sh
* 0.27s for ./index.sh

Using yq (python version):
* 0m17.991s for ./check_mandatory_fields.sh
* 0m15.440s for ./index.sh

So... I'm actually thinking we should NOT use the yq python wrapper. :(

check_mandatory_fields.sh
```diff
@@ -28,12 +28,12 @@ function check_field() {
...
 for i in $(ls -1 "${devfilesDir}/"*"/meta.yaml" | sort); do
     echo "Checking devfile '${i}'"
-    id=$(grep "displayName:" "${i}" | sed -e "s#^displayName: ##" -e 's#"##g')
+	id=$(yq ."${displayName}" "$i" | tr -d "\"") # trim quotes from the field value
     full_id=${id}:${i}
 
     unset NULL_OR_EMPTY_FIELDS
     for field in "${metaInfoFields[@]}"; do
-      value=$(grep "${field}:" "${i}" | sed -e "s#^${field}: ##" -e 's#"##g')
+	  value=$(yq ."${field}" "$i" | tr -d "\"") # trim quotes from the field value
```
nickboldt: I don't actually think we should do this, I'm only showing this change to demonstrate how grep and yq are nearly the same syntax. I refute the argument that sed is less legible than yq for trivial yaml lookups like these.

nickboldt:
out of date.

PR #79, opened 2019-08-13
Add image with golang 1.12 for plugin broker devfile.

AndrienkoAleksandr:
### What does this PR do?
Add image with golang 1.12 for plugin broker devfile.

https://github.com/eclipse/che/issues/14126

### DIFF
--- a/arbitrary-users-patch/base_images
+++ b/arbitrary-users-patch/base_images
@@ -1,6 +1,7 @@
 che-python-3.6          centos/python-36-centos7:1
 che-php-7               eclipse/php:7.1-che7
 che-golang-1.10         golang:1.10.7-stretch
+che-golang-1.12         golang:1.12-stretch
 che-java11-gradle       gradle:5.2.1-jdk11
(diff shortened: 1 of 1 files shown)


arbitrary-users-patch/base_images
```diff
@@ -1,6 +1,7 @@
 che-python-3.6          centos/python-36-centos7:1
 che-php-7               eclipse/php:7.1-che7
 che-golang-1.10         golang:1.10.7-stretch
+che-golang-1.12         golang:1.12-stretch
```
amisevsk: Since our current golang devfile is just used for the one project, we should ensure `golang:1.12-stretch` works with https://github.com/golang/example.git and replace `che-golang-1.10` with `che-golang-1.12`. I'd rather not build more images than is necessary here.

amisevsk:
Nightly CI failed: https://ci.centos.org/job/devtools-che-devfile-registry-build-master/

Created https://quay.io/repository/eclipse/che-golang-1.12

PR #80, opened 2019-08-13
Replace mikefarah/yq with kislyuk/yq

amisevsk:
### What does this PR do?
Replaces [`mikefarah/yq`](https://github.com/mikefarah/yq) with [`kislyuk/yq`](https://github.com/kislyuk/yq) and reworks `index.sh` and `check_mandatory_fields.sh` to accommodate this change.

- `index.sh` is greatly simplified and can basically be done in one step (with a workaround for self-links)
- `check_mandatory_fields.sh` no longer uses yq at all, since a simple grep command is sufficient to check if the fields are not present or empty.

Since timing is a concern in eclipse/che#14076, 
```
$ time ./check_mandatory_fields.sh devfiles
real	0m 0.05s
user	0m 0.03s
sys	0m 0.01s
$ time ./index.sh > /build/devfiles/index.json
real	0m 0.26s
user	0m 0.21s
sys	0m 0.04s
```

It's some minor work from testing https://github.com/eclipse/che/issues/14076

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,8 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #
-FROM mikefarah/yq as builder
-RUN apk add --no-cache bash
+FROM alpine:3.10 AS builder
+RUN apk add --no-cache py-pip jq bash && pip install yq
 
 COPY .htaccess README.md *.sh /build/
 COPY /devfiles /build/devfiles
--- a/check_mandatory_fields.sh
+++ b/check_mandatory_fields.sh
@@ -12,42 +12,21 @@ set -e
 
 FIELDS=('displayName' 'description' 'tags' 'icon' 'globalMemoryLimit')
 
-
-
-# check that field value, given in the parameter, is not null or empty
-function check_field() {
-  if [[ $1 == "null" || $1 = "" ]];then
-    return 1;
-  fi
-  return 0
-}
-
-readarray -d '' arr < <(find "$1" -name 'meta.yaml' -print0)
-
-for i in "${arr[@]}"
-do
-    id=$(yq r "$i" displayName | sed 's/^"\(.*\)"$/\1/')
-    full_id=${id}:${i}
-
-    echo "Checking devfile '${i}'"
-
+readarray -d '' metas < <(find devfiles -name 'meta.yaml' -print0)
+for meta in "${metas[@]}"; do
+    echo "Checking devfile '${meta}'"
     unset NULL_OR_EMPTY_FIELDS
-
-    for FIELD in "${FIELDS[@]}"
-    do
-      VALUE=$(yq r "$i" "$FIELD")
-
-      if ! check_field "${VALUE}";then
-        NULL_OR_EMPTY_FIELDS+="$FIELD "
-      fi
+    for field in "${FIELDS[@]}"; do
+        if ! grep -q "^${field}:.*\S" $meta; then
+            NULL_OR_EMPTY_FIELDS+="$field "
+        fi
     done
-
     if [[ -n "${NULL_OR_EMPTY_FIELDS}" ]];then
-      echo "!!!   Null or empty mandatory fields in '${full_id}': $NULL_OR_EMPTY_FIELDS"
-      INVALID_FIELDS=true
+        echo "!!!   Null or empty mandatory fields in ${meta}: $NULL_OR_EMPTY_FIELDS"
+        INVALID_FIELDS=true
     fi
 done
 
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -6,8 +6,8 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #
-FROM mikefarah/yq as builder
-RUN apk add --no-cache bash
+FROM alpine:3.10 AS builder
```
nickboldt: if you're going to rebase to a new base image, why not use ubi-minimal? https://github.com/nickboldt/containers/blob/master/ubi8-httpd-yq/Dockerfile

check_mandatory_fields.sh
```diff
@@ -12,42 +12,21 @@ set -e
 
 FIELDS=('displayName' 'description' 'tags' 'icon' 'globalMemoryLimit')
```
nickboldt: This array var is different from the nearly-identical approach in the plugin registry. Could we use the same variables in both repos?

nickboldt:
Does your greatly improved index.sh solution work for plugin registry too? that'd be much more awesome since that's a more complex build.

Dockerfile
amisevsk: Could be, I went with alpine mostly because I meant this change to be as transparent as possible. `mikefarah/yq` is based off alpine, so this is the most similar. 

Will look into moving to ubi8-minimal.

check_mandatory_fields.sh
amisevsk: I'd keep them separate mostly since their purposes are different and there's no requirement to sync metadata fields between plugins and devfiles. I'm mostly viewing these are loop variables, but I'll look at the registry as well and see it makes sense.

amisevsk:
@nickboldt I'm looking at the registry next, I don't think it'll be too difficult. 

One note about this PR -- it does make one change to the registry: it adds
```
links:
  self: /devfiles/<folder>/meta.yaml
```
to each meta.yaml. This *shouldn't* be a breaking change, but I haven't tested this.

Dockerfile
amisevsk: After glancing at the additional configuration/workarounds required, I'd rather do the ubi8 switch in a separate PR.

benoitf:
it's so bad to use python with so much libs in the image

amisevsk:
> it's so bad to use python with so much libs in the image

@benoitf Why? Could you elaborate? Why is `mikefarah/yq` better? It's certainly less capable.

As I see it, the python `yq` wrapper is the standard tool to do the job we want to do; it is used by many members on the team and has a syntax that is familiar to most developers (who have used `jq`). The dependencies, also, are not really obscure python libraries -- `yq` depends on `xmltodict`, `toml`, `setuptools`, and `PyYAML`. The mikefarah version seems to only have been used because there is a pre-built docker image.

The resulting images are of identical size.

benoitf:
> @benoitf Why? Could you elaborate? Why is mikefarah/yq better? It's certainly less capable.

>As I see it, the python yq wrapper is the standard tool to do the job we want to do; it is used by >many members on the team and has a syntax that is familiar to most developers (who have used >jq). The dependencies, also, are not really obscure python libraries -- yq depends on xmltodict, >toml, setuptools, and PyYAML. The mikefarah version seems to only have been used because >there is a pre-built docker image.

I would say that "it is used by many members on the team" is not so true as you're replacing the usage of mikefarah in at least 2 repositories so before it was more used.

> The resulting images are of identical size.
true but false as well.
Mario wants to have scripts in the runtime so it means having yq in the runtime image
So it will result in 50MB+ image that has just dummy shell script 👎

amisevsk:
> I would say that "it is used by many members on the team" is not so true as you're replacing the usage of mikefarah in at least 2 repositories so before it was more used.

The reason I first encountered `yq` was it was required for something when I started working on Che (and I remember leaving comments about the `yq` implementation confusion back then). Plus, `yq` is just `jq` that accepts yaml files, so it's more familiar. I'll concede that it may not be as common now, though.

> Mario wants to have scripts in the runtime so it means having yq in the runtime image
So it will result in 50MB+ image that has just dummy shell script

I'm also against putting the scripts in the final image (and have PRs that remove the current copying as well). I think the discussion in https://github.com/eclipse/che/issues/14115 will need to be reconsidered since the offline caching steps are implemented slightly differently from the original. In particular, I believe the purpose of this is to support

> INSTALLERS - Add offline option to helm and che-operator. If the option is set `set_plugins_url.sh` should be executed as soon as the plugin registry is started.

which I'm not clear on as a goal.

PR #90, opened 2019-08-28
Tag patched images

amisevsk:
### What does this PR do?
- Adds ability update devfiles at build time to use custom patched images (e.g. you can use `docker.io/my-organization/che-java-maven:my-tag` instead of the current `quay.io/eclipse/che-java-maven:nightly`)
    - This is enabled via build args on the Dockerfiles, that come with defaults. 
- Updates the CICO release script to use this functionality when building releases of the devfile registry. Now, the release script will
    1. Build images with the arbitrary-user-id patch, tagged with release version (from `VERSION` file)
    2. Build devfile registry with devfiles using these patched images.

These changes should also make it easier to configure downstream registries, if e.g. we want to build `ubi8` or `rhel` versions of patched images and push them to a different quay org.

**Note**: there are detailed commit messages for each commit explaining their changes.

### Alternative approaches considered
We could remove the automatic updating during build and instead require devfiles to be updated manually, but I think this approach is better because
1. It is more automated and avoids adding a step to our release process
2. It makes it easier to maintain a forked devfile registry; if e.g. someone wanted to cherry-pick changes into a release branch (or private fork) they would need to resolve merge conflicts and/or re-patch devfiles.

I'm open to reworking this PR to go with this approach if people agree that this is better, though.

https://github.com/eclipse/che/issues/14165


build.sh
```diff
@@ -8,4 +8,14 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-docker build -t quay.io/eclipse/che-devfile-registry:nightly .
+VERSION=$(head -n 1 VERSION)
+case $VERSION in
+  *SNAPSHOT)
+    echo "Building plugin registry"
+    # docker build -t quay.io/eclipse/che-devfile-registry:nightly .
```
ibuziuk: commented out accidentally?

build.sh
```diff
@@ -8,4 +8,14 @@
...
+    ;;
+  *)
+    echo "Building plugin registry for release ${VERSION}"
+    # docker build -t quay.io/eclipse/che-devfile-registry:nightly . --build-arg "PATCHED_IMAGES_TAG=${VERSION}"
```
ibuziuk: commented out accidentally?

build.sh
```diff
@@ -8,4 +8,14 @@
 # SPDX-License-Identifier: EPL-2.0
 #
 
-docker build -t quay.io/eclipse/che-devfile-registry:nightly .
+VERSION=$(head -n 1 VERSION)
+case $VERSION in
+  *SNAPSHOT)
+    echo "Building plugin registry"
```
ibuziuk: maybe add info that for SNAPSHOT versions 'nightly' version of plugin runtime containers will be used?

build.sh
```diff
@@ -8,4 +8,14 @@
...
+    # docker build -t quay.io/eclipse/che-devfile-registry:nightly .
+    ;;
+  *)
+    echo "Building plugin registry for release ${VERSION}"
```
ibuziuk: maybe add info that for the release version  `RELEASE` version of plugin runtime containers will be used?

ibuziuk:
In general, the idea with patching looks really good :+1:

cico_functions.sh
```diff
@@ -108,13 +108,27 @@ function build_and_push() {
...
+    --build-arg PATCHED_IMAGES_ORG=${ORGANIZATION} \
+    --build-arg PATCHED_IMAGES_TAG=${TAG}
+  echo "CICO: release '${TAG}' version of devfile registry built"
+  tag_push "${REGISTRY}/${ORGANIZATION}/${IMAGE}:${TAG}"
+  echo "CICO: release '${TAG}' version of devfile registry pushed to '${REGISTRY}/${ORGANIZATION}' organization"
+}
+
+# Build images patched to work on OpenShift (work with arbitrary user IDs) and push
+# them to registry
+function build_patched_base_images() {
```
ibuziuk: Just double-checking - This will be executed only for `centos` target right? Since push with `rhel` is going to fail

ibuziuk: or most like during the `release` CI images would be build & pushed twice to the public quay.io `eclipse` registry

update_devfile_patched_image_tags.sh
```diff
@@ -0,0 +1,32 @@
...
+
+DEFAULT_TAG="nightly"
+DEFAULT_REGISTRY="quay.io"
+DEFAULT_ORGANIZATION="eclipse"
+
+TAG=${TAG:-$DEFAULT_TAG}
+REGISTRY=${REGISTRY:-$DEFAULT_REGISTRY}
+ORGANIZATION=${ORGANIZATION:-$DEFAULT_ORGANIZATION}
+
+if [ "$TAG" = "$DEFAULT_TAG" ] && [ "$ORGANIZATION" = "$DEFAULT_ORGANIZATION" ] && [ "$REGISTRY" = "$DEFAULT_REGISTRY" ]; then
```
ibuziuk: what happens if TAG = GIT Commit hash. Wondering about `master` CI

cico_functions.sh
amisevsk: It would execute during rhel builds as well -- I suppose more attention should be given to this case; it's confusing since we're trying to support four flows through the same code based off environment variables.

When the release script is called with `TARGET='rhel'`, the changes in this PR would 
1. Build arbitrary-user-id patched base images for `quay.io/openshiftio/<image>:<release-tag>` (these repos don't currently exist and so the build **would fail**)
2. Build a devfile registry `quay.io/openshiftio/rhel-che-devfile-registry:<release>` that references `quay.io/openshiftio/<image>:<release>`

For nightly `rhel`-target builds, the build should be unmodified from previous.

update_devfile_patched_image_tags.sh
amisevsk: I'm not sure `TAG` is ever set to the commit hash. In the CI scripts, I only see it set to either `nightly` or `head -n 1 VERSION`. We do tagging based off the git commit as well, but it is just a [tag on the built image](https://github.com/eclipse/che-devfile-registry/pull/90/files#diff-3961cf6dd52815a90fe783caf9af804dR103) and is only done for regular builds (i.e. ones where we don't change devfiles). 

Note that `update_devfile_patched_image_tags.sh` will only execute when `docker build` is called with `--build-arg PATCHED_IMAGES_[REG|ORG|TAG]=<non-default>`.

cico_functions.sh
amisevsk: I could rework the CI build to use the `eclipse` org for all patched images (so that rhel builds reference `quay.io/eclipse/<image>:<release>`) but that would involve more complicated logic. 

Currently I've tried to structure the CI so that once the environment variables are set, the same `REPOSITORY`, `ORGANIZATION`, and `TAG` are used for both patched images and the built registry.

amisevsk:
I've added a commit to separate out dockerfile scripts from CI scripts to hopefully make the structure clearer (move dockerfiles and their scripts to `./build/scripts` `./build/dockerfiles`). This commit also modifies the build to not copy unnecessary files (scripts, etc.) into the final docker image.

update_devfile_patched_image_tags.sh
ibuziuk: TAG based on the commit hash is set as part of every CI https://github.com/eclipse/che-devfile-registry/blob/master/cico_functions.sh#L100-101 but I guess it will not be called for regular CI case since --build-arg are not used here - https://github.com/eclipse/che-devfile-registry/blob/master/cico_functions.sh#L116-L117

ibuziuk:
@l0rd @skabashnyuk @rhopp @vparfonov could you please take a look?

update_devfile_patched_image_tags.sh
amisevsk: To add, the tag based on commit has *is* set, but it uses the env var [`${GIT_COMMIT_TAG}`](https://github.com/eclipse/che-devfile-registry/blob/53d89c445b47be6cb2431ca34ebbec5df4cba377/cico_functions.sh#L68), not `${TAG}`, so even in the release case the git commit hash would not be used.

amisevsk:
Rebased to resolve merge conflict.

amisevsk:
@l0rd @skabashnyuk @rhopp @vparfonov Please review if you have a moment, I'd like to get this PR merged (or closed if it's not what we want).

ibuziuk:
@amisevsk I believe we can merge now ;-)

PR #98, opened 2019-09-22
combine steps to eliminate needless...

nickboldt:
combine steps to eliminate needless layers

Change-Id: Id65fbb03cc8762a2cb6d9f2db09e3e5da05cf26c

### DIFF
--- a/arbitrary-users-patch/Dockerfile
+++ b/arbitrary-users-patch/Dockerfile
@@ -2,11 +2,8 @@ ARG FROM_IMAGE
 FROM ${FROM_IMAGE}
 USER 0
 # Set permissions on /etc/passwd and /home to allow arbitrary users to write
-RUN mkdir -p /home/user
-RUN chgrp -R 0 /home
-RUN chmod -R g=u /etc/passwd /home
 COPY [--chown=0:0] entrypoint.sh /
-RUN chmod +x /entrypoint.sh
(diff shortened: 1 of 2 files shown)


arbitrary-users-patch/Dockerfile
```diff
@@ -2,11 +2,8 @@ ARG FROM_IMAGE
 FROM ${FROM_IMAGE}
 USER 0
 # Set permissions on /etc/passwd and /home to allow arbitrary users to write
-RUN mkdir -p /home/user
-RUN chgrp -R 0 /home
-RUN chmod -R g=u /etc/passwd /home
 COPY [--chown=0:0] entrypoint.sh /
-RUN chmod +x /entrypoint.sh
+RUN mkdir -p /home/user && chgrp -R 0 /home && chmod -R g=u /etc/passwd /home && chmod +x /entrypoint.sh
```
benoitf: just a small note, it seems chmod is optional if file in git has +x permission already no ?

amisevsk: This is true, but also it currently is `rw-r--r--`

nickboldt: and it doesn't hurt to document in the dockerfile the permissions you want, rather than assuming the permissions in the filesystem will be used/respected/are correct.

PR #106, opened 2019-09-27
Don't update registry and org in RHEL release builds

amisevsk:
### What does this PR do?
Fixes an issue in the CI: since https://github.com/eclipse/che-devfile-registry/pull/105 disables building patched base images for RHEL-target release builds, we need to also not update the devfiles to point at the RHEL registry/organization. Otherwise, devfiles in the RHEL version of the repo will point at the `quay.io/openshiftio/<images>/<tag>` images that are no longer built.

Release CI

### DIFF
--- a/cico_functions.sh
+++ b/cico_functions.sh
@@ -116,8 +116,6 @@ function build_and_push() {
 function build_and_push_release() {
   echo "CICO: building release '${TAG}' version of devfile registry"
   docker build -t ${IMAGE} -f ${DOCKERFILE_PATH} . \
-    --build-arg PATCHED_IMAGES_REG=${REGISTRY} \
-    --build-arg PATCHED_IMAGES_ORG=${ORGANIZATION} \
     --build-arg PATCHED_IMAGES_TAG=${TAG}
   echo "CICO: release '${TAG}' version of devfile registry built"
   tag_push "${REGISTRY}/${ORGANIZATION}/${IMAGE}:${TAG}"


cico_functions.sh
```diff
@@ -115,10 +115,17 @@ function build_and_push() {
...
   echo "CICO: building release '${TAG}' version of devfile registry"
-  docker build -t ${IMAGE} -f ${DOCKERFILE_PATH} . \
-    --build-arg PATCHED_IMAGES_REG=${REGISTRY} \
-    --build-arg PATCHED_IMAGES_ORG=${ORGANIZATION} \
-    --build-arg PATCHED_IMAGES_TAG=${TAG}
+  if [ "$TARGET" == "rhel" ]; then
+    # Do not update registry and tags on RHEL releases, since we don't build
+    # patched images there.
+    docker build -t ${IMAGE} -f ${DOCKERFILE_PATH} . \
+      --build-arg PATCHED_IMAGES_TAG=${TAG}
```
ibuziuk: @amisevsk just to confirm, for rhel build the following argument will be used by default?

> ARG PATCHED_IMAGES_REG="quay.io"
> ARG PATCHED_IMAGES_ORG="eclipse"

Maybe we can simply remove ` if [ "$TARGET" == "rhel" ]` and override only tag in general ?

amisevsk: That would be another option; the reason it's currently done the way it is is to support the case of building a registry top to bottom for your own org. I guess that's less useful in our specific CI though, since really only we would use it.

amisevsk:
I took your suggestion and removed the org and registry params. Please double check.

ibuziuk:
@amisevsk I like that one more ;)

PR #108, opened 2019-10-03
Use java11 language tool for stacks with java11.

AndrienkoAleksandr:
### What does this PR do?
We have two stacks with java11 images, but development tools used with [java8 version](https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/redhat/java/latest/meta.yaml#L16). This pr fix it.

none

### DIFF
--- a/devfiles/java-gradle/devfile.yaml
+++ b/devfiles/java-gradle/devfile.yaml
@@ -11,7 +11,7 @@ projects:
 components:
   -
     type: chePlugin
-    id: redhat/java/latest
+    id: redhat/java11/latest
   -
     type: dockerimage
     alias: gradle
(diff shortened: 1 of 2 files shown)


svor:
@AndrienkoAleksandr The sample which is using for these devfiles was specified to use 1.8 compiler, you can check it here (https://github.com/che-samples/console-java-simple/blob/master/pom.xml#L11). After changing java version in plugin container, LS will show warning `The compiler compliance specified is 1.8 but a JRE 11 is used`. So probably we need to update our sample.

AndrienkoAleksandr:
@svor I prepared branch with sample upgraded to java 11: https://github.com/che-samples/console-java-simple/pull/12/files and I am awaiting review and merge from @sleshchenko or @amisevsk - they have permission to merge. Then I will create pr to use source sample with java 11.

PR #110, opened 2019-10-06
Allow overriding container images used in devfiles

amisevsk:
### What does this PR do?
Allow overriding image registry, organization, and tag in devfiles at runtime using the env vars
```
CHE_DEVFILE_IMAGE_REGISTRY_URL
CHE_DEVFILE_IMAGE_REGISTRY_ORGANIZATION
CHE_DEVFILE_IMAGE_REGISTRY_TAG
```

Also updates helm charts and OpenShift template.

https://github.com/eclipse/che/issues/14693


deploy/kubernetes/che-devfile-registry/templates/deployment.yaml
```diff
@@ -56,3 +56,6 @@ spec:
             memory: {{ .Values.cheDevfileRegistryMemoryLimit }}
           requests:
             memory: 256Mi
+        envFrom:
+          - configMapRef:
+              name: che-devfile-registry
```
davidfestal: Seems to me that we should make this config map ref `optional`, to avoid update problems.

amisevsk: Done.

build/dockerfiles/Dockerfile.rhel
```diff
@@ -41,9 +42,10 @@ RUN chmod a+rwX            /run/httpd
...
 COPY .htaccess README.md /var/www/html/
 COPY --from=builder /build/devfiles /var/www/html/devfiles
-USER 0
-RUN chmod -R g+rwX /var/www/html/devfiles
 
 STOPSIGNAL SIGWINCH
 
-ENTRYPOINT ["/usr/sbin/httpd", "-D", "FOREGROUND"]
+COPY --from=builder /build/devfiles /var/www/html/devfiles
+COPY ./build/dockerfiles/entrypoint.sh /usr/bin/
```
nickboldt: after copying file over again we don't have to fix permissions?

build/dockerfiles/Dockerfile.rhel
```diff
@@ -41,9 +42,10 @@ RUN chmod a+rwX            /run/httpd
...
+ENTRYPOINT ["/usr/bin/entrypoint.sh"]
+CMD ["/usr/sbin/httpd", "-D", "FOREGROUND"]
```
nickboldt: we could put this INSIDE the entrypoint.sh script. What benefit to being separate? Also you can have the same entrypoint check for the alpine/ubuntu vs. rhel path to run httpd:

```
# start httpd
if [[ -x /usr/sbin/httpd ]]; then /usr/sbin/httpd -D FOREGROUND
elif [[ -x /usr/bin/run-httpd ]]; then /usr/bin/run-httpd
fi
```

(You might recognize that from https://github.com/eclipse/che-plugin-registry/blob/master/scripts/entrypoint.sh )

build/dockerfiles/Dockerfile.rhel
amisevsk: The file in the repo has 755 permissions so it's not an issue. I can add the line for explicitness if you prefer, though.

build/dockerfiles/Dockerfile.rhel
amisevsk: The benefit of having the `httpd` command as a CMD instead of in the ENTRYPOINT is that it's easier to override for debugging:
```
docker run -it --rm <registry> /bin/bash
```

This is generally how it's done AFAIK, e.g. for `registry.access.redhat.com/rhscl/httpd-24-rhel7:2.4-104` we have
```
"Entrypoint": ["container-entrypoint"],
"Cmd": ["/usr/bin/run-httpd"],
```

As for the actual command being run, I'm copying what was done previously (as an entrypoint instead of a command). I can add the wrapper script from plugin registry if necessary, though.

amisevsk:
> Suggest embedding the logic to start httpd inside the entrypoint script, like we do for the plugin registry, because consistency ;)

You just [approved](https://github.com/eclipse/che-plugin-registry/pull/245/files#diff-b9a489dd6d57174e10e145befa2a4034R116) a commit that moves the start httpd logic into a CMD ;)

build/dockerfiles/Dockerfile.rhel
amisevsk: Added `rhel.entrypoint.sh` script to match plugin registry structure.

build/dockerfiles/Dockerfile.rhel
amisevsk: Added chmod run to set entrypoint permissions.

amisevsk:
@nickboldt Updated PR, please take another look.

build/dockerfiles/Dockerfile.rhel
nickboldt: thanks I just like being explicit because sometimes you pull files on a disrespectful OS (or your GH settings are wack) and you accidentally commit a 644 file when it should be 755. Or so I've heard, it's NEVER happened to me :)

build/dockerfiles/Dockerfile.rhel
nickboldt: but you can:

`docker run -it --rm --entrypoint /bin/bash <container> -c "ls -la ${HOME}"` too, regardless of the stated ENTRYPOINT or CMD, right?

(key is to change the order of the command and explicitly use `--entrypoint /bin/bash` BEFORE the container `reg/org/image:tag`)

build/dockerfiles/Dockerfile.rhel
```diff
@@ -41,9 +42,10 @@ RUN chmod a+rwX            /run/httpd
...
-USER 0
-RUN chmod -R g+rwX /var/www/html/devfiles
 
 STOPSIGNAL SIGWINCH
 
-ENTRYPOINT ["/usr/sbin/httpd", "-D", "FOREGROUND"]
+COPY ./build/dockerfiles/rhel.entrypoint.sh ./build/dockerfiles/entrypoint.sh /usr/local/bin/
+RUN chmod g+rwX /usr/local/bin/entrypoint.sh /usr/local/bin/rhel.entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+CMD ["/usr/local/bin/rhel.entrypoint.sh"]
```
nickboldt: you're still inconsistent with the Dockerfile approach. My thought was to have fewer differences between Dockerfile and rhel.Dockerfile so maintaining them both would be easier. Wasn't just to nest the command for the sake of nesting... apologies if that wasn't clear/obvious.

build/dockerfiles/rhel.entrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "${USER_NAME:-jboss}:x:$(id -u):0:${USER_NAME:-jboss} user:${HOME}:/sbin/nologin" >> /etc/passwd
```
nickboldt: we only need this in RHEL, not alpine/ubuntu?

amisevsk: The registry has been working fine without it since its inception; unless there's a compelling case for including it without knowing why, I'd rather leave it out.

I'm against pushing rhel-related complexity into the default image for the sake of consistency.

build/dockerfiles/Dockerfile.rhel
amisevsk: I view this as consistent between the two; we already have to so much additional work in the rhel dockerfile because rhel, and I view `rhel.entrypoint.sh` as another facet of that problem.

Tacking the run-httpd script onto entrypoint.sh is pointless; the entrypoint should work for both images and do the devfile-registry specific setup that needs to happen before starting. The command can do the container specific starting of the Apache server, in a container specific way (since rhel containers demand weirdness).

In my view: 

- dockerfile entrypoint = application specific startup requirements (rewrite meta.yamls)
- dockerfile cmd = start whatever process we need in the container

build/dockerfiles/Dockerfile.rhel
amisevsk: Yeah but that's not the typical way docker images are written -- there's a reason you need an additional flag. In fact, that's not the way other centos httpd container implementations are written.

Also, your form makes it harder to test that the entrypoint is doing what we expect -- to test the container you would have to 1) override entrypoint and start with `/bin/bash`, 2) run the entrypoint from bash.

build/dockerfiles/rhel.entrypoint.sh
nickboldt: yeah turns out this was a mistake... the container runs as root so this jboss user thing is not needed.

build/dockerfiles/Dockerfile.rhel
nickboldt: OK works for me.

build/dockerfiles/rhel.entrypoint.sh
amisevsk: It may still be needed, but I'd rather add it when we know it's necessary. On OpenShift, the way things work is that containers are started running as a (mostly random) UID. As far as I know, the things that are impacted by this are:
- File permissions: you basically can't rely user-level permissions in container files, since the UID is inconsistent across deployments; any files that you need read/write/execute need to be owned by the root group and have those permissions there
- Username dependency: Since your UID is generated at startup, you won't have a username or home directory set up; this can cause problems for userspace programs that rely on e.g. storing things in the home directory, or anything that depends on being able to access a username. I don't *think* this impacts Apache, since it doesn't touch user files. The containers we're using already have appropriate permissions for everything Apache needs. 

The second part is what the line above is designed to address (so that a username is available). We need it on Che workspace images, since there we're basically emulating a user terminal/IDE and various apps there depend on a username, home directory, login shell, etc.

FWIW also, I believe OpenShift 4.x clusters can be configured to automatically provision an entry in `/etc/passwd` at container startup using cri-o, but I haven't played with it too much.

PR #112, opened 2019-10-07
Build an offline version of the Devfile registry image

amisevsk:
**Important Note:** this PR is set to target the `override-images` branch until https://github.com/eclipse/che-devfile-registry/pull/110 is merged, as it also requires modification of the container entrypoint. Once PR #110 is merged, I will update target branch to master.

### What does this PR do?
Enables building an offline version of the devfile registry.

As in the similar feature for the plugin registry, this is done by specifying `--target offline-registry` during the docker build (with the current online registry being built by `--target registry`).

The offline registry downloads all project zip files for projects hosted on github and stores them in the `/resources` dir. Devfiles are updated to contain a placeholder, which is filled from an env var (`CHE_DEVFILE_HTTPS_ENDPOINT`) that must be set at runtime.

https://github.com/eclipse/che/issues/14733

### Additional Info
Tested on OpenShift 3.11, manually setting the required environment variable. It works generally as desired, except I encountered issue https://github.com/eclipse/che/issues/14800.

### Caveat

For this solution to be viable, all the sample projects to be bundled need to have their license/copyright info correctly configured.

See https://issues.jboss.org/browse/CRW-317?focusedCommentId=13798222&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13798222 for a list of the problems with the sample projects' licenses/copyright info.


amisevsk:
Update: the last commit adds an attempt to determine the registry's service IP and port from k8s env vars when the required environment variable is not set. This *should* work in many cases but is rather fragile. I haven't tested it for workspaces in another namespace, etc.

amisevsk:
Squashed and rebased onto master now that #110 is merged; PR is ready for review

cc: @l0rd @nickboldt

build/dockerfiles/entrypoint.sh
```diff
@@ -16,13 +16,20 @@
 # By default, this script will operate on the `/var/www/html/devfiles` directory.
 # This can be overridden by the environment variable $DEVFILES_DIR
 #
+# In addition, this script will perform the necessary set up for the offline
+# devfile registry, replacing placeholders in all devfiles based off environment
+# variable
+#     CHE_DEVFILE_HTTPS_ENDPOINT
```
davidfestal: So I assume that in airgap mode, from the operator, we would also have to add this variable in the new devfile registry config map, based on the external URL of the devfile registry ?

@l0rd do you confirm ?

cc @tomgeorge

davidfestal:
@l0rd @nickboldt  So I assume that if we would like to use the devfile registry *with* embedded samples inside for the airgap mode, we should first have built that with the `offine-registry` mode.
So we should have 2 different registry images and switch from one to the other in the operator if we use airgap ?
Not sure I understand how this will be used in the case of the operator airgap mode.

nickboldt:
I suspect for CRW 2 we will only **build & publish the offline/airgap option**. Yes, it's a bigger image to download than the online flavour, but it's also more guaranteed to work as all the plugins/projects will be self-contained in the devfile and plugin reg containers.

l0rd:
I agree with @nickboldt. It will be the default downstream (and we should consider to make it the default upstream as well c.f. [this comment](https://github.com/eclipse/che/issues/14733#issuecomment-539049640https://github.com/eclipse/che/issues/14733#issuecomment-539049640))

l0rd:
@amisevsk I think we should update the container build section in the README file as well and the ["Customizing the devfile and plugin-in registries" section](https://www.eclipse.org/che/docs/che-7/customizing-the-devfile-and-plug-in-registries/) in the doc. But considered that this PR may block other PRs we can merge it and address the doc in a separate one.

davidfestal:
> we should consider to make it the default upstream as well

+1

build/dockerfiles/entrypoint.sh
amisevsk: Yes, that's accurate; we need the environment variable to be able to rewrite devfiles to point at exposed endpoints. Without the environment variable we do try to use k8s-provisioned env vars to determine service IP and route, but I haven't tested this much and it may be unreliable.

amisevsk:
@l0rd I added some docs to the README.md and opened issue https://github.com/eclipse/che/issues/14866 to track the larger documentation task

build/dockerfiles/entrypoint.sh
l0rd: @davidfestal yes that's the weakness of this solution. There is no reliable way to figure the value automatically.

amisevsk: In particular, without the env var set, we try to go off of 
```
URL="http://${CHE_DEVFILE_REGISTRY_SERVICE_HOST}:${CHE_DEVFILE_REGISTRY_SERVICE_PORT}"
```

which depends on the devfile registry server being called `che-devfile-registry`, and may not work in some scenarios (I don't know the internals of networking on OpenShift).

amisevsk:
Rebased on master to fix merge conflict.

amisevsk:
> For this solution to be viable, all the sample projects to be bundled need to have their license/copyright info correctly configured.

Che issue for this: https://github.com/eclipse/che/issues/14790

build/dockerfiles/rhel.Dockerfile
```diff
@@ -96,3 +96,13 @@ ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 CMD ["/usr/local/bin/rhel.entrypoint.sh"]
 
 # append Brew metadata here
+
```
nickboldt: tiny nitpick but the `# append Brew metadata here` should be at the end of the file :)

amisevsk: fixed

PR #115, opened 2019-10-10
switch to 100% RHEL-basd build (not Alpine...

nickboldt:
switch to 100% RHEL-basd build (not Alpine build + RHEL deployment)

Change-Id: I414a172f5b3b4e93ef613a5f4dca3d68c69c43a6


build/dockerfiles/rhel.Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+          popd >/dev/null; \
+        fi; \
+      done; \
+      chmod -c +x /usr/local/bin/*; \
+    else \
+      /usr/bin/pip3.6 install yq jsonschema; \
+    fi && \
+    ln -s /usr/bin/python3.6 /usr/bin/python && \
+    # test install worked
+    for d in python yq jq jsonschema; do echo -n "$d: "; $d --version; done
```
amisevsk: Down the road (i.e. not in this PR) we should think about moving this into a script or something.

It's also worth noting that `RUN` in a dockerfile will default to using `/bin/sh`, so the above script only works in this case since `/bin/sh` on the UBI image is `/bin/bash`.

build/scripts/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -x
+
+# start httpd
+if [[ -x /usr/sbin/httpd ]]; then /usr/sbin/httpd -D FOREGROUND
+#elif [[ -x /usr/bin/run-httpd ]]; then /usr/bin/run-httpd
+fi
```
amisevsk: Sorry, I merged #110 before looking at this PR. The master branch of repo now contains `build/dockerfiles/entrypoint.sh`, so we should use/modify that instead.

build/dockerfiles/rhel.Dockerfile
```diff
@@ -0,0 +1,118 @@
...
+    -e "s,AllowOverride None,AllowOverride All," && \
+    chmod a+rwX /etc/httpd/conf /run/httpd
+STOPSIGNAL SIGWINCH
+# END these steps might not be required
+
+RUN mkdir /var/www/html/devfiles
+COPY .htaccess README.md /var/www/html/
+COPY --from=builder /build/devfiles /var/www/html/devfiles
+RUN chmod -R g+rwX /var/www/html/devfiles
+COPY /build/scripts/*entrypoint.sh /usr/local/bin/
```
amisevsk: See `build/dockerfiles/entrypoint.sh` and `build/dockerfiles/rhel.entrypoint.sh` (merged in #110). I think `rhel.entrypoint.sh` covers the functionality of `uid_entrypoint` and `entrypoint` from this PR.

build/scripts/list_containers.sh
```diff
@@ -0,0 +1,38 @@
...
+#
+# Copyright (c) 2019 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+# pull all external references to container images
```
amisevsk: I'm not clear on the purpose of this script, could you clarify? As it stands it looks like it only does
```
yq -r '.components[].image | strings' **/devfile.yaml | sort | uniq
```

build/dockerfiles/rhel.Dockerfile
nickboldt: done!

build/scripts/list_containers.sh
nickboldt: yeah, good call. yq script pushed instead of lengthy bash thing.

build/scripts/list_containers.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# $ ./build/scripts/list_containers.sh devfiles/
+
+set -e
+
+if [[ ! $1 ]]; then DIR=$(dirname "$0"); else DIR="$1"; fi
+
+# search in devfiles folder, eg., $1 = devfiles/
+echo "BEGIN list of external containers in $DIR folder:"
+cd $DIR; yq -r '.components[].image | strings' **/devfile.yaml | sort | uniq
```
amisevsk: `cd` is unnecessary, you could just do `$DIR/**/devfile.yaml`. 

If you want to maintain the indentation from the previous version, you could also pipe to sed: `| sed "s/^/    /g"`

PR #118, opened 2019-10-12
must also chmod /etc/httpd/logs/ to allow...

nickboldt:
must also chmod /etc/httpd/logs/ to allow httpd to write logs

Change-Id: If88709ff09f3cb07ffced30e78652bcd194c4d52

### DIFF
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -81,7 +81,7 @@ RUN sed -i /etc/httpd/conf/httpd.conf \
     -e "s,logs/error_log,/dev/stderr," \
     -e "s,logs/access_log,/dev/stdout," \
     -e "s,AllowOverride None,AllowOverride All," && \
-    chmod a+rwX /etc/httpd/conf /run/httpd
+    chmod a+rwX /etc/httpd/conf /run/httpd /etc/httpd/logs/
 STOPSIGNAL SIGWINCH
 # END these steps might not be required
 


build/dockerfiles/rhel.Dockerfile
```diff
@@ -81,7 +81,7 @@ RUN sed -i /etc/httpd/conf/httpd.conf \
     -e "s,logs/error_log,/dev/stderr," \
     -e "s,logs/access_log,/dev/stdout," \
```
amisevsk: My understanding is that these lines configure httpd to log to stderr/stdout instead of a logfile.

build/dockerfiles/rhel.Dockerfile
```diff
@@ -81,7 +81,7 @@ RUN sed -i /etc/httpd/conf/httpd.conf \
...
     -e "s,AllowOverride None,AllowOverride All," && \
-    chmod a+rwX /etc/httpd/conf /run/httpd
+    chmod a+rwX /etc/httpd/conf /run/httpd /etc/httpd/logs/
```
amisevsk: Looking at the image we're currently using for the rhel devfile registry, I see
```
$ ls -al /etc/httpd
drwxr-xr-x. 1 root root  32 Sep 11 18:24 .
drwxr-xr-x. 1 root root  19 Oct 15 12:45 ..
drwxrwxrwx. 1 root root  24 Oct 15 12:45 conf
drwxrwxrwx. 1 root root  22 Oct 15 12:45 conf.d
drwxrwxrwx. 2 root root 244 Sep 11 18:24 conf.modules.d
lrwxrwxrwx. 1 root root  16 Sep 11 18:24 logs -> /var/log/httpd24
drwxr-xr-x. 4 root root  48 Sep 11 18:24 modsecurity.d
lrwxrwxrwx. 1 root root  44 Sep 11 18:24 modules -> /opt/rh/httpd24/root/usr/lib64/httpd/modules
lrwxrwxrwx. 1 root root  34 Sep 11 18:24 run -> /opt/rh/httpd24/root/var/run/httpd
lrwxrwxrwx. 1 root root  34 Sep 11 18:24 state -> /opt/rh/httpd24/root/var/lib/httpd

$ ls -al /var/log/httpd24
total 0
drwxrwx---. 1 default    root 54 Oct 15 12:45 .
drwxr-xr-x. 1 root       root 21 Sep 11 18:24 ..
-rw-r-----. 1 1000450000 root  0 Oct 15 12:45 modsec_audit.log
-rw-r-----. 1 1000450000 root  0 Oct 15 12:45 modsec_debug.log
```
so it would appear that this change wouldn't affect anything.

build/dockerfiles/rhel.Dockerfile
nickboldt: well, then maybe on rhel they're not working.

build/dockerfiles/rhel.Dockerfile
nickboldt: It does. Without this the image crashes on startup as httpd can't write to the logs folder. With this change, the registries work.

nickboldt: ok that's so weird. Maybe running the image locally is fine but in Minishift it failed for me :( Can't find the specific tag in Quay that was crashlooping on Friday.

PR #122, opened 2019-10-17
che #14916 Release CI should  build image based on the git commit hash (required for Hosted Che release cycle)

ibuziuk:
### What does this PR do?
che #14916 Release CI should  build an image based on the git commit hash (required for Hosted Che release cycle)

https://github.com/eclipse/che/issues/14916

### DIFF
--- a/cico_functions.sh
+++ b/cico_functions.sh
@@ -110,14 +110,19 @@ function build_and_push() {
   fi
 }
 
-# Build release version of devfile registry, using ${TAG} as a tag. For release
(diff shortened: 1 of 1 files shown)


ibuziuk:
This PR should be also cherry-picked to the 7.3.x / release branch, otherwise we will not be able to update the registry on dsaas

ibuziuk:
@amisevsk nightly does not build the images from the release / 7.x branches, so we have to explicitly build the GIT_COMMIT_HASH for releases (otherwise we will not be able to update registries on prod-preview / prod)

PR #126, opened 2019-10-18
Deprecate che-golang-1.10 in favor of che-golang-1.12

amisevsk:
### What does this PR do?
Remove `che-golang-1.10` from list of arbitrary UID patched images in favor of `che-golang-1.12`. This PR will stop nightly/release builds of that image and make new workspaces created using to go devfile use 1.12. We still need to decide what to do with the existing `che-golang-1.10` images on quay.io, since old workspaces may depend on it.

https://github.com/eclipse/che/issues/14265

### DIFF
--- a/arbitrary-users-patch/base_images
+++ b/arbitrary-users-patch/base_images
@@ -1,6 +1,5 @@
 che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7
 che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
-che-golang-1.10         golang:1.10.7-stretch
 che-golang-1.12         golang:1.12-stretch
 che-java11-gradle       gradle:5.2.1-jdk11
 che-java11-maven        maven:3.6.0-jdk-11
--- a/devfiles/go/devfile.yaml
+++ b/devfiles/go/devfile.yaml
@@ -18,7 +18,7 @@ components:
 -
   type: dockerimage
   # this version is used in the plugin
-  image: quay.io/eclipse/che-golang-1.10:nightly
+  image: quay.io/eclipse/che-golang-1.12:nightly
   alias: go-cli
   env:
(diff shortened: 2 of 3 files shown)


nickboldt:
Two thoughts here
* For CRW 2 we'll be using golang 1.11 as there's no 1.12 in RHCC
* You said "deprecate" but then you deleted stuff. That sounds more like "remove" than "deprecate". :) Where's the grace period? (FWIW, I like the idea of just removing the old stuff, but then you should rename this issue to "remove" rather than "deprecate". :) )

amisevsk:
It doesn't remove it per se, but does remove it from the nightly build; the image will still be available to be referenced by old workspaces. With the changes, new workspaces would automatically use 1.12. I wouldn't call this a "remove", since remove implies that `che-golang-1.10` will no longer be available. 

In terms of deprecation procedure, we don't currently have one for these images. We'd need a way to notify users of old golang workspaces that we're no longer building the image, and I don't know a good way of doing that. Part of the issue is that the devfile registry only serves devfiles for workspace creation; already-existing workspaces store their own devfiles and so the registry is completely disconnected from the process.

@nickboldt WDYT about adding a `#deprecated` to the image in `base_images`? My main goal here is to signal to people that we should avoid using 1.10 unless necessary, since maintaining both 1.12 and 1.10 is weird.

nickboldt:
I'm good with you deleting it completely from here.

Maybe you could add a note on the registry's image listing, like we did here:

https://quay.io/repository/eclipse-che/che7-java11-maven?tab=info ??

Then it's labelled at the source, and the build script/process doesn't have to put `#deprecated` on stuff that has died and won't be coming back.

ibuziuk:
@amisevsk could you please add deprecation note to the https://quay.io/repository/eclipse/che-golang-1.10

amisevsk:
@ibuziuk Done: https://quay.io/repository/eclipse/che-golang-1.10?tab=info

PR #128, opened 2019-10-23
Add list of external images to server when offline registry is built

amisevsk:
### What does this PR do?
Adds building a list of externally-referenced images to the offline build to make it easier for users to know which images need to be pulled into a private registry.

Required for https://github.com/eclipse/che/issues/14866

Also rewrites the `build.sh` script to be similar to the one in the plugin registry (with options)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,21 +1,6 @@
-#!/bin/sh
-#
-# Copyright (c) 2012-2018 Red Hat, Inc.
-# This program and the accompanying materials are made
-# available under the terms of the Eclipse Public License 2.0
-# which is available at https://www.eclipse.org/legal/epl-2.0/
-#
-# SPDX-License-Identifier: EPL-2.0
-#
-# Build Che devfile registry image. Note that this script will read the version
-# in ./VERSION; if it is *-SNAPSHOT, devfiles in the registry will use nightly-tagged
-# images with the arbitrary user IDs patch. If ./VERSION contains otherwise,
-# the devfiles in the registry will instead use the value in ./VERSION.
-#
-
 #!/bin/bash
 #
-# Copyright (c) 2012-2018 Red Hat, Inc.
+# Copyright (c) 2019 Red Hat, Inc.
 # This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
 # which is available at https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -13,16 +13,107 @@
...
+else
+    echo ""
+    docker build \
+        -t "$IMAGE" \
+        -f ./build/dockerfiles/Dockerfile \
+        --build-arg LATEST_ONLY="${LATEST_ONLY}" \
+        --target registry .
+fi
+
+IMAGE="${REGISTRY}/${ORGANIZATION}/che-devfile-registry:${TAG}"
```
nickboldt: you're doing this step twice? see line 89

build.sh
```diff
@@ -13,16 +13,107 @@
...
+        esac
+    done
+}
+
+parse_arguments "$@"
+
+IMAGE="${REGISTRY}/${ORGANIZATION}/che-plugin-registry:${TAG}"
+echo -n "Building image '$IMAGE' "
+if [ "$OFFLINE" = true ]; then
+    echo "in offline mode"
```
nickboldt: why not simpler...

```
if [ "$OFFLINE" = true ]; then TARGET="offline-registry"; fi
docker build \
        -t "$IMAGE" \
        -f ./build/dockerfiles/Dockerfile \
        --build-arg LATEST_ONLY="${LATEST_ONLY}" \
        --target ${TARGET} .
...
```

build.sh
```diff
@@ -13,16 +13,107 @@
 # the devfiles in the registry will instead use the value in ./VERSION.
 #
 
+#!/bin/bash
```
nickboldt: seems an odd place to put a sh-bang. shouldn't this be on line 1? Oh, wait there's already one there. Not sure what's going on here.... but -1.

build.sh
amisevsk: Yeah, this is a bad copy-paste error. Fixed.

build.sh
amisevsk: Copy paste error from plugin registry build.sh. Fixed.

build.sh
amisevsk: Yup, copy paste error. Fixed.

amisevsk:
@nickboldt Sorry for the confusion, that build.sh script was a big ol' copy paste error from plugin registry. Should be fixed now.

PR #133, opened 2019-10-28
che 14947: Releasing 7.3.1 version of the devfile registry

ibuziuk:
### What does this PR do?
Cherry-picking all commits from `master` to `7.3.x`
Releasing 7.3.1 version of the devfile registry

https://github.com/eclipse/che/issues/14947


nickboldt:
+1, I think all these changes from master are already downstream in CRW's version of devfile registry so when we switch over to use 7.3.1 it should be seamless.

PR #135, opened 2019-10-31
Add devfile with java sample based on google cloud sample 

svor:
### What does this PR do?
Adds new devfile with [java-guestbook](https://github.com/svor/java-guestbook) sample. It is a java web application with Spring Boot and MongoDB which is based on [Google Cloud Sample](https://github.com/GoogleCloudPlatform/cloud-code-samples/tree/master/java/java-guestbook) and can be deployed on k8s cluster.

https://github.com/eclipse/che/issues/14651

### DIFF
--- a/devfiles/java-mongo/meta.yaml
+++ b/devfiles/java-mongo/meta.yaml
@@ -1,0 +1,6 @@
+---
+displayName: Java with Spring Boot and MongoDB
+description: Java stack with OpenJDK 8, MongoDB and Spring Boot Guestbook demo application
+tags: ["Java", "OpenJDK", "Maven", "Spring Boot", "MongoDB"]
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+globalMemoryLimit: 3796Mi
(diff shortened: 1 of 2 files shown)


devfiles/java-mongo/meta.yaml
```diff
@@ -0,0 +1,6 @@
+---
+displayName: Java with Spring Boot and MongoDB
+description: Java stack with OpenJDK 8, MongoDB and Spring Boot Guestbook demo application
+tags: ["Java", "OpenJDK", "Maven", "Spring Boot", "MongoDB"]
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+globalMemoryLimit: 3284Mi
```
amisevsk: A memory limit of `3284Mi` will prevent this devfile from running on che.openshift.io. Is there any way to reduce the required memory below `3Gi`?

tsmaeder: Spring boot pulls in a boatload of maven dependencies. Is using it a choice or are there good reasons for it?

svor: @amisevsk I set 300Mi for mongo container and 512 for maven, tested it and looks like it works without problems. So, now global memory limit is 2884Mi.

amisevsk: :+1:

PR #144, opened 2019-11-11
Add Quarkus devfile

svor:

### What does this PR do?
Adds a devfile to support Qurakus tools.
#### This devfile includes

- 2 components:
  - dev component is based on `quay.io/quarkus/centos-quarkus-maven:19.2.1` which contains Java 8,  Maven 3.6.0 and GraalVM 19.2.1
  - qurkus plugin (for more information take a look at https://github.com/eclipse/che-plugin-registry/pull/291)  

- 2 commands:
  - mvn package
  - mvn compile quarkus:dev

- The application could be debugged by predefined debug configuration in case `mvn compile quarkus:dev` was executed before.

This devfile doesn't include commands to run native compilation because it requires more than 3GB RAM for workspace in total and it's not available on Hosted Che.   

https://github.com/eclipse/che/issues/14506

### DIFF
--- a/devfiles/quarkus/devfile.yaml
+++ b/devfiles/quarkus/devfile.yaml
@@ -1,0 +1,69 @@
+---
+apiVersion: 1.0.0
+metadata:
+  generateName: quarkus-
+projects:
+  -
+    name: quarkus-quickstarts
+    source:
+      type: git
+      location: "https://github.com/quarkusio/quarkus-quickstarts.git"
+      sparseCheckoutDir: /getting-started/
+components:
+  -
+    type: chePlugin
+    id: redhat/quarkus/latest
+  -
+    type: dockerimage
+    alias: maven
+    image: quay.io/eclipse/che-quarkus:nightly
+    env:
+      - name: JAVA_OPTS
+        value: "-XX:MaxRAMPercentage=50.0 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10
+          -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90
+          -Dsun.zip.disableMemoryMapping=true -Xms20m -Djava.security.egd=file:/dev/./urandom
+          -Duser.home=/home/user"
+      - name: MAVEN_OPTS
+        value: $(JAVA_OPTS)  
+    memoryLimit: 512Mi
+    mountSources: true
+    volumes:
+      - name: m2
+        containerPath: /home/user/.m2
+    endpoints:
+      - name: 'quarkus-getting-started'
+        port: 8080    
+commands:
+  -
+    name: maven package
+    actions:
+      -
+        type: exec
+        component: maven
+        command: "mvn package"
+        workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
+  -
+    name: maven run in development mode
+    actions:
+      -
+        type: exec
+        component: maven
+        command: "mvn compile quarkus:dev"
+        workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
+  -
+    name: Attach remote debugger
+    actions:
+    - type: vscode-launch
+      referenceContent: |
+        {
+          "version": "0.2.0",
+          "configurations": [
+            {
+              "type": "java",
+              "request": "attach",
+              "name": "Attach to Remote Quarkus App",
+              "hostName": "localhost",
+              "port": 5005
+            }
+          ]
+        }
--- a/devfiles/quarkus/devfile.yaml
+++ b/devfiles/quarkus/devfile.yaml
@@ -31,11 +31,15 @@ components:
       - name: m2
         containerPath: /home/user/.m2
     endpoints:
-      - name: 'quarkus-getting-started'
+      - name: 'quarkus-development-server'
         port: 8080    
+      - name: 'hello-greeting-endpoint'
+        port: 8080
+        attributes:
+          path: /hello/greeting/che-user
 commands:
   -
-    name: maven package
+    name: Package the application
     actions:
       -
         type: exec
@@ -43,7 +47,7 @@ commands:
         command: "mvn package"
         workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
   -
-    name: maven run in development mode
+    name: Start development server
     actions:
       -
         type: exec
--- a/devfiles/quarkus/devfile.yaml
+++ b/devfiles/quarkus/devfile.yaml
@@ -12,7 +12,7 @@ projects:
 components:
   -
     type: chePlugin
-    id: redhat/quarkus/latest
+    id: redhat/quarkus-java8/latest
   -
     type: dockerimage
     alias: maven
--- a/devfiles/quarkus/meta.yaml
+++ b/devfiles/quarkus/meta.yaml
@@ -1,0 +1,6 @@
+---
+displayName: Quarkus Tools
+description: Quarkus Tools with OpenJDK 8 and Maven 3.6.0
+tags: ["Java", "Quarkus", "OpenJDK", "Maven", "Debian"]
+icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+globalMemoryLimit: 2674Mi


devfiles/quarkus/devfile.yaml
```diff
@@ -0,0 +1,68 @@
...
+      location: "https://github.com/quarkusio/quarkus-quickstarts.git"
+      sparseCheckoutDir: /getting-started/
+components:
+  -
+    type: chePlugin
+    id: redhat/quarkus/latest
+  -
+    type: dockerimage
+    alias: maven
+    image: quay.io/eclipse/che-java8-maven:nightly
```
sunix: we should include the ability to run the native compilation to this devfile

svor: could you please clarify why do we need that?

sunix: With Quarkus, you have a way to produce a tiny native executable https://quarkus.io/guides/building-native-image

for instance:
https://github.com/sunix/che-quarkus-demo/blob/8f4f4fa8e09158853e9cd900c0fe93cac18da484/devfile.yaml#L109-L121

svor: looks like GraalVM Native Image generation requires at least 2Gi RAM to perform a native image build. So, if we give this value + 1,5Gi for Quarkus VS Code plugin (Java LS, Quarkus LS, Java Debugger) container, it'd be more than 3Gi - it is too much for che.openshift.io

tsmaeder:
> will it be compatible with Hosted Che (3GB ram)?
modulo native compilation, it should

svor:
@ibuziuk except building a native executable, it should

devfiles/quarkus/devfile.yaml
sunix: To me we have to produce a devfile for che. Not che.openshift.io.

ibuziuk: upstream devfile registry is used on Hosted Che, so all the devfiles in the registry must be compatible with Hosted Che. Otherwise, we need to come up with Hosted Che specific devfile registry (smth. that should be avoided)

sunix:
Not including native executable compilation would show only 1/3 of the value of Che for quarkus developers. IMHO

devfiles/quarkus/devfile.yaml
sunix: > looks like GraalVM Native Image generation requires at least 2Gi RAM to perform a native image build. So, if we give this value + 1,5Gi for Quarkus VS Code plugin (Java LS, Quarkus LS, Java Debugger) container, it'd be more than 3Gi - it is too much for che.openshift.io

Could you give a try in a single container with vscode ext for java/quarkus and running on a container with graalvm ?

svor:
It is possible to execute native compilation but it'll be failed with OOM on hosted Che, that's why I didn't add commands into devfile. If the user will have more OOM for Che it could add such commands into the workspace or just execute them in the dev terminal

sunix:
@ibuziuk 
> upstream devfile registry is used on Hosted Che, so all the devfiles in the registry must be compatible with Hosted Che. Otherwise, we need to come up with Hosted Che specific devfile registry (smth. that should be avoided)

If the goal of hosted Che is to show the value of Che to potential/possible users, for a Quarkus devfile, we need to have Quarkus + native compilation. If we don't have that, there is no value compared to a VSCode or another IDE. We need to reconsider the 3Go limitation and give 4 or 5Gigs at least if we cannot optimize it.

ibuziuk:
> We need to reconsider the 3Go limitation and give 4 or 5Gigs at least if we cannot optimize it.

isn't it smth you have already raised on the planning? This is a PM decision, so, for now, all the devfiles should be compatible with the 3Gb RAM limits available on Hosted Che - https://www.eclipse.org/che/docs/che-7/hosted-che/#terms-of-service_hosted-che

amisevsk:
IMO native compilation is a deploy step most of the time (with the caveat there are some minor incompatibilities). If I was developing a quarkus app, I my flow would be something like
- Develop using `quarkus:dev` for hot code reload and testing/debugging
- Once ready to deploy, run a builder separately to create and test the native binary.

Native compilation in the workspace is a "nice to have" but not essential, since generally native compilation is something you would do once, at the end of preparing a PR.

devfiles/quarkus/devfile.yaml
```diff
@@ -0,0 +1,69 @@
...
+          -Dsun.zip.disableMemoryMapping=true -Xms20m -Djava.security.egd=file:/dev/./urandom
+          -Duser.home=/home/user"
+      - name: MAVEN_OPTS
+        value: $(JAVA_OPTS)  
+    memoryLimit: 512Mi
+    mountSources: true
+    volumes:
+      - name: m2
+        containerPath: /home/user/.m2
+    endpoints:
```
amisevsk: Consider also adding an endpoint for `/hello/greeting/che-user` or similar, to test an application endpoint.

devfiles/quarkus/devfile.yaml
```diff
@@ -0,0 +1,69 @@
...
+  -
+    name: maven package
+    actions:
+      -
+        type: exec
+        component: maven
+        command: "mvn package"
+        workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
+  -
+    name: maven run in development mode
```
amisevsk: In my opinion, command names should say "what you're doing" rather than "how it's being done", so maybe `Start development server` is clearer here.

Up to you, though -- I'm fine with either.

devfiles/quarkus/devfile.yaml
```diff
@@ -0,0 +1,69 @@
...
+          -Duser.home=/home/user"
+      - name: MAVEN_OPTS
+        value: $(JAVA_OPTS)  
+    memoryLimit: 512Mi
+    mountSources: true
+    volumes:
+      - name: m2
+        containerPath: /home/user/.m2
+    endpoints:
+      - name: 'quarkus-getting-started'
```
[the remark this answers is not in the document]
svor: changed!

devfiles/quarkus/devfile.yaml
svor: Added one more endpoint

devfiles/quarkus/devfile.yaml
svor: Updated

sunix:
> IMO native compilation is a deploy step most of the time (with the caveat there are some minor incompatibilities). If I was developing a quarkus app, I my flow would be something like
> 
> * Develop using `quarkus:dev` for hot code reload and testing/debugging
> * Once ready to deploy, run a builder separately to create and test the native binary.
> 
> Native compilation in the workspace is a "nice to have" but not essential, since generally native compilation is something you would do once, at the end of preparing a PR.

Yes this is nice to have but essential to show the value of Che: this is what is hard to have if you are working on a Quarkus app on a traditional Desktop environment. The current devfile is easy to setup with VSCode + quarkus ext... so what is the point to use Che ?

nickboldt:
Looks like all of Angel's concerns have been addressed; which only leaves @ibuziuk 's question of 3G vs. 4-5G as the required memory footprint.

tsmaeder:
Folks, we should not make merging this PR dependent on the native compilation commands. In the short run, we cannot increase memory on hosted Che. In the short run, we cannot run a separate command line registry for hosted che. 
IMO, we have 3 courses of action open:
1. Merge the PR "as is" 
2. Wrap native compilation with a script that graciously fails when memory is lacking (aka on hosted che) and run that as the native compile command
3. Don't add Quarkus support.

@slemeur make a decision, please.

nickboldt:
My $0.02 based on existing CRW 2.1 requirements...

#3 isn't an option  - Quarkus support is required for CRW 2.1, and being able to dogfood it in some way in Che 7.7 and in Hosted Che is IMHO a valuable way to ensure it's working downstream too.

#2 seems like the best approach - failing gracefully is an important part of UX when running on disparate clusters. Error message on failure should include detailed call to action so workspace user knows they need to ask for more RAM from their cluster administrator. I would further tell the user that if they don't have access to their administrator, this functionality won't be available (eg., if they need 5G in Hosted Che, they're going to have a bad time.)

tsmaeder:
I would go ahead and merge this PR as is. We can add native compile commands in the next sprint.

slemeur:
+1 on the approach

PR #147, opened 2019-11-20
Che4z 1.0.0

filipkroupa:
### What does this PR do?
Make Che4z devfile available in devfile registry

N/A

[https://www.eclipse.org/che/docs/che-7/eclipse-che4z/](https://www.eclipse.org/che/docs/che-7/eclipse-che4z/)

Signed-off-by: Filip Kroupa filip.kroupa@broadcom.com


devfiles/che4z/devfile.yaml
```diff
@@ -0,0 +1,73 @@
...
+    id: bitlang/cobol/latest
+  - 
+    type: chePlugin
+    id: broadcommfd/explorer-for-endevor/latest
+  - 
+    id: eclipse/che-machine-exec-plugin/7.3.0
+    type: chePlugin
+  - 
+    id: eclipse/che-theia/7.3.0
+    type: cheEditor
```
amisevsk: Devfiles should not contain specific versions of `che-theia` and `che-machine-exec-plugin`, since that overrides Che automatically using the latest release (see [che.properties](https://github.com/eclipse/che/blob/6e0ce6d101b337e0d77ffa1b25f5b0bcd621ce06/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L558-L571))

devfiles/che4z/devfile.yaml
```diff
@@ -0,0 +1,73 @@
...
+           apiVersion: v1
+           kind: PersistentVolumeClaim
+           metadata:
+            name: zowe-claim
+           spec:
+            accessModes:
+             - ReadWriteOnce
+            resources:
+              requests:
+                storage: 128Mi
```
amisevsk: Please add trailing newline.

devfiles/che4z/devfile.yaml
```diff
@@ -0,0 +1,73 @@
...
+    type: chePlugin
+  - 
+    id: eclipse/che-theia/7.3.0
+    type: cheEditor
+  - 
+    alias: zowe-cli
+    type: kubernetes
+    mountSources: true
+    referenceContent: |
+         kind: List
```
amisevsk: Is it possible to rewrite the devfile to follow the existing structure, or is the Kubernetes List required?

devfiles/che4z/devfile.yaml
```diff
@@ -0,0 +1,73 @@
...
+           metadata:
+            name: zowe
+           spec:
+            volumes:
+             - name: zowe
+               persistentVolumeClaim:
+                 claimName: zowe-claim
+            containers:
+             -
+              image: 'ca-docker-brightside-docker.bintray.io/cli-mainframe-basic-stack:2019-11-19T15-56-35'
```
amisevsk: This docker image, on OpenShift, could run into issues around user IDs. See [the arbitrary user id patch](https://github.com/eclipse/che-devfile-registry/tree/master/arbitrary-users-patch) we run for other workspace images.

devfiles/che4z/devfile.yaml
```diff
@@ -0,0 +1,73 @@
...
+              env:
+               - name: THEIA_PLUGINS
+                 value: local-dir:///plugins
```
amisevsk: I'm not clear on why this env var is here, since this container does not mount `/plugins`

devfiles/che4z/devfile.yaml
filipkroupa: Added missing newline

devfiles/che4z/devfile.yaml
filipkroupa: We are aware of this, but we prefer to release with version we tested against. It happened in the past that our extension stopped working in next release of che-theia, for example because version of Theia framework changed. This does not affect user's ability to switch che-theia to 'latest' in workspace configuration.

filipkroupa:
I'm afraid we cannot provide simple test project to try out all plugins functionality since it requires server counterparts which currently cannot be exposed outside of intranet. To test the Cobol LSP functionality there is for example project [Marbles-Lite-Brightside-Samples](https://github.com/CATechnologies/Marbles-Lite-Brightside-Samples) which should contain valid IBM Cobol samples.

devfiles/che4z/devfile.yaml
MarkAckert: THEIA_PLUGINS was set because the docker container also bundles a the Zowe Explorer VSCode extension, deployed via the theia-remote library. Longer term we would like to separate the VSCode extension into a separate plugin container and use `chePlugin`, but the first pass was bundled due to colocation of runtime and filesystem dependencies. The Zowe Explorer was built for a user workstation, and it's undergoing enhancements in parallel to better support Che+Theia deployment.

devfiles/che4z/devfile.yaml
MarkAckert: We saw the user ID issue on OpenShift, and will look into arbitrary user ID support. 

Is OpenShift support required at this point in time?

devfiles/che4z/devfile.yaml
MarkAckert: Could you expand on what you refer to with 'existing structure'?  The Kubernetes List was required when compared to `dockerimage` type deployment, but I'm not sure if this is what you're referring to. We have persistence requirements for Zowe CLI profiles which required an additional PVC.

devfiles/che4z/devfile.yaml
amisevsk: That makes sense to me :+1:

devfiles/che4z/devfile.yaml
amisevsk: In the past, using a Kubernetes List was necessary because the dockerimage format didn't support everything you might need, but now it's possible to specify volumes, environment vars, endpoints, etc. on a dockerimage component (e.g. for [java-spring](https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/java-web-spring/devfile.yaml#L17-L35)).

I personally don't have a strong preference either way, but going this route would match the other devfiles.

devfiles/che4z/devfile.yaml
amisevsk: It's preferable, in my opinion, but if it doesn't interfere with the functionality you need it's not strictly necessary. It does mean, however, that anything that depends on username or `$HOME` directories will behave weirdly. 

We can also potentially add the image `ca-docker-brightside-docker.bintray.io/cli-mainframe-basic-stack:2019-11-19T15-56-35` to the arbitrary user ID patch set -- I can provide steps if you would like to go down that path (basically you would just need to add a line to [base_images](https://github.com/eclipse/che-devfile-registry/blob/master/arbitrary-users-patch/base_images)).

devfiles/che4z/devfile.yaml
amisevsk: Makes sense to me, so the Zowe Explorer runs in the `ca-docker-brightside-docker.bintray.io/cli-mainframe-basic-stack:2019-11-19T15-56-35` image instead of in the Theia container?

MarkAckert: That's correct

devfiles/che4z/devfile.yaml
MarkAckert: If there's no objection - I'd like to keep the devfile as kubernetes for this release and revisit the `dockerimage` format for a future revision. I think there's value in pursuing the `dockerimage` format in addition to alignment with the other devfiles. Which release of Che added the env, volumes, etc. support?

devfiles/che4z/devfile.yaml
MarkAckert: >  It does mean, however, that anything that depends on username or $HOME directories will behave weirdly.

Verified this was the case, the Zowe Explorer extension expects the container `WORKDIR` to be set to `/projects`, and the arbitrary user ID results in `WORKDIR` `/`. This results in a failure to load properly. The rest of the stack software works as expected.

I am pretty sure the arbitrary user ID patch set would create issues, so I tried to make some "quick" changes to our docker images using the patch set as a template. Interestingly, the Zowe Explorer extension fails to load at first, but works properly after a browser refresh. I think we need to spend a little more time digging into the details of our images before we can support the arbitrary user ID, but I have no problem going down that path.

devfiles/che4z/devfile.yaml
amisevsk: That works for me, it's not like the referenceContent approach has any issues I can see -- it's mostly a consistency thing. I believe the additional fields (env, volumes, etc.) were in [the 7.0.0 release](https://github.com/eclipse/che/blob/7.0.0/wsmaster/che-core-api-workspace/src/main/resources/schema/devfile.json#L444) but our documentation somewhat lacking until relatively recently. 

For reference, the current documentation on devfiles is [here](https://www.eclipse.org/che/docs/che-7/making-a-workspace-portable-using-a-devfile/#component-type-dockerimage_making-a-workspace-portable-using-a-devfile). If you do decide to go down the path of using a `dockerimage` component, it'd be great to document any issues you encounter along the way.

filipkroupa:
We are targeting [Milestone 7.5.0](https://github.com/eclipse/che/milestone/105) Can we expect this to be part of the release?

amisevsk:
> We are targeting Milestone 7.5.0 Can we expect this to be part of the release?

Merged for 7.5.0, thanks again @filipkroupa

PR #148, opened 2019-11-25
Remove quotes.

monaka:
Signed-off-by: Masaki Muranaka <monaka@monami-ya.com

### What does this PR do?

Resolves eclipse/che#15293.

eclipse/che#15293


nickboldt:
why would your plugin need multiple icons? AFAIU only one can be displayed in the dashboard.

amisevsk:
@nickboldt 
> why would your plugin need multiple icons? AFAIU only one can be displayed in the dashboard.

The issue occurs if two plugins specify separate images, and is a pretty big mistake on my part in initially implementing it. We haven't seen it yet since *every* plugin uses the same Che logo.

monaka:
@amisevsk I agree #150 is the better approach. I close this now and move to #150.

PR #149, opened 2019-11-25
Prepare an image for Quarkus dev environment

svor:

### What does this PR do?
Patches quay.io/quarkus/centos-quarkus-maven:19.2.1 to use it Quarkus devfile as a dev environment.

**This image contains:**
OpenJDK 1.8.0_232
GraalVM 19.2.1
Apache Maven 3.6.1 

https://github.com/eclipse/che/issues/14506

### DIFF
--- a/arbitrary-users-patch/base_images
+++ b/arbitrary-users-patch/base_images
@@ -10,3 +10,4 @@ che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
 che-php-7               eclipse/php:7.1-che7
 che-python-3.6          centos/python-36-centos7:1
 che-python-3.7          python:3.7.4-slim
+che-quarkus             quay.io/quarkus/centos-quarkus-maven:19.2.1


arbitrary-users-patch/base_images
```diff
@@ -10,3 +10,4 @@ che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
 che-php-7               eclipse/php:7.1-che7
 che-python-3.6          centos/python-36-centos7:1
 che-python-3.7          python:3.7.4-slim
+che-qurkus-19.2.1       quay.io/quarkus/centos-quarkus-maven:19.2.1
```
amisevsk:
```suggestion
che-quarkus-19.2.1       quay.io/quarkus/centos-quarkus-maven:19.2.1
```

amisevsk:
Note also that before merging this PR, repo `quay.io/eclipse/che-quarkus-19.2.1` must be created

arbitrary-users-patch/base_images
amisevsk: Additionally, consider just using `che-quarkus` or `che-quarkus-19`; if we name it after a bugfix release we will have to start a new repo for `19.2.2`, etc.

I'd prefer the images we push to quay.io to refer to compatible sets of versions, rather than specific targetted versions.

svor: I'v changed it to **che-quarkus**. 19.2.1 - it is version of GraalVM, so probably for now it makes sense to name it che-quarkus.

svor:
> Note also that before merging this PR, repo quay.io/eclipse/che-quarkus-19.2.1 must be created

@amisevsk Thanks, I'll ask to create the repo before merging

arbitrary-users-patch/base_images
amisevsk: :+1:

svor:
@ibuziuk  It'll be possible to run quarkus devfile with 3GB. It won't be possible to use native compilation, because for that we need to have > 2Gb

nickboldt:
Looks OK to me but I'd prefer if the image was based on UBI than Centos, because then we'd be able to more easily reuse it in downstream. :)

amisevsk:
@svor Created https://quay.io/repository/eclipse/che-quarkus

svor:
> @svor Created https://quay.io/repository/eclipse/che-quarkus

@amisevsk thanks!

PR #150, opened 2019-11-25
Fix caching of images when multiple images specified

amisevsk:
### What does this PR do?
Fix bug reading list of images specified in meta.yamls, where more than one unique URL would cause caching script to fail.

https://github.com/eclipse/che/issues/15293

### DIFF
--- a/build/scripts/cache_images.sh
+++ b/build/scripts/cache_images.sh
@@ -40,7 +40,9 @@ while read -r image; do
   image_dir="${RESOURCES_DIR%/}/${image_dir%/*}"
   mkdir -p "$image_dir"
 
-  cached_image="${image_dir%/}/${filename%\?*}"
+  # Strip query and fragment components from image URL
+  cached_image="${image_dir%/}/${filename%%\?*}"
+  cached_image="${cached_image%%\#*}"
   mv "$file" "$cached_image"
   echo "  Downloaded image $image to $cached_image"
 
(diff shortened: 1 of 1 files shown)


amisevsk:
@monaka WDYT about this PR vs yours (https://github.com/eclipse/che-devfile-registry/pull/148)? They both work but if we ever get around to enabling CI on this repo, shellcheck will fail on the missing quotes in #148.

amisevsk:
I added a commit to resolve another issue I ran into while testing:

Images can be passed with query parameters, e.g. https://raw.githubusercontent.com/redhat-developer/codeready-workspaces/master/dependencies/che-devfile-registry/images/type-java.svg?sanitize=true

The devfile registry currently will save images with filename `type-java.svg?sanitize=true`, which makes them inaccessible in offline mode. Commit 3c68339ca91a887f1bd835f9ca580a174b96677e fixes that.

build/scripts/cache_images.sh
```diff
@@ -40,13 +40,13 @@ for image in "${images[@]}"; do
   image_dir="${RESOURCES_DIR%/}/${image_dir%/*}"
   mkdir -p "$image_dir"
 
-  cached_image="${image_dir%/}/${filename}"
+  cached_image="${image_dir%/}/${filename%\?*}"
```
monaka: I guess `#` fragment also should be cared here. @amisevsk

amisevsk: Thanks @monaka, I missed that. I've fixed it to strip both query and fragment components, and also to avoid an edge case where a url contains multiple `?` or `#` chars.

monaka: I checked 2c34856 and LGTM.

amisevsk: Thanks @monaka!

PR #160, opened 2019-12-09
Unset MAVEN_CONFIG env in Maven based images

svor:

### What does this PR do?
This PR unsets MAVEN_CONFIG env to make it possible to run **mvnw** scripts. 

https://github.com/eclipse/che/issues/13926


arbitrary-users-patch/build_images.sh
```diff
@@ -31,7 +33,11 @@ while read -r line; do
   base_image_name=$(echo "$line" | tr -s ' ' | cut -f 1 -d ' ')
   base_image=$(echo "$line" | tr -s ' ' | cut -f 2 -d ' ')
   echo "Building ${NAME_FORMAT}/${base_image_name}:${TAG} based on $base_image ..."
-  docker build -t "${NAME_FORMAT}/${base_image_name}:${TAG}" --no-cache --build-arg FROM_IMAGE="$base_image" "${SCRIPT_DIR}"/
+  script_dir=${SCRIPT_DIR}
+  if [[ ${base_image:0:5} = ${MAVEN_BASE} ]] ; then
```
nickboldt: This seems like a too-restrictive check. What if the base image name is just a java image which happens to have maven in it? https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-devfile-registry/devfiles/02_java-maven/devfile.yaml#L18

Why not do a check like this?

```
  # substring check for "java" or "maven" in ${base_image}
  if [[ ${base_image} = *"maven"* ]] || [[ ${base_image} = *"java"* ]]; then 
```

nickboldt:
Should we do a similar change to the devfile.yaml files for these?

https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-devfile-registry/devfiles/00_java-eap-maven/devfile.yaml#L20-L21
https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-devfile-registry/devfiles/01_java-eap-thorntail/devfile.yaml#L20-L21
https://github.com/redhat-developer/codeready-workspaces/blob/master/dependencies/che-devfile-registry/devfiles/02_java-maven/devfile.yaml#L20-L21

And do I need to apply `ENV MAVEN_CONFIG=` in this dockerfile?

http://pkgs.devel.redhat.com/cgit/containers/codeready-workspaces-stacks-java/tree/Dockerfile?h=crw-2.0-rhel-8#n82

arbitrary-users-patch/build_images.sh
amisevsk: I continue to be strongly against this sort of patch in the arbitrary user patch section. If we want an image with `MAVEN_CONFIG` unset then we should build one explicitly.

amisevsk: Also worth noting that this workaround is *specific* to the [dockerhub community maven image](https://github.com/carlossg/docker-maven/blob/master/jdk-8/mvn-entrypoint.sh#L45), and is not necessary in e.g. an image where you install maven yourself.

svor:
@amisevsk My point was to unset `MAVEN_CONFIG` for both images `che-java11-maven` and `che-java8-maven` in one place. Probably I agree with you that it wasn't the best choice to patch them in a general script by adding maven specific logic. 
So, if it's OK for you to unset `MAVEN_CONFIG` in all devfiles where we are using `che-java11-maven` and `che-java8-maven`, I'll do that

nickboldt:
What if we built a pair of small, UBI 8 based images that work on OpenShift and include either JDK 8 or 11?

Then you would not be dependent on Alpine, would get the benefits of CVE fixes, and would provide something that could more closely be used for downstream too.

Also the JDK 11 image could be the basis for quarkus devfiles.

amisevsk:
I'm happy with either option
- Updating devfiles makes sense to me; we're using a particular image for the maven dependency, so the devfiles have to have a workaround for that particular image.
- We can update the image we use to avoid having to unset `MAVEN_CONFIG`, by building a ubi-based image.

Both also have (minor) downsides:
- Updating our devfiles works but might be a sticking point for users trying to reuse our "che-maven" image (they would have to figure out the same fix for their devfile, if they don't use one of ours as a base). 
- It's not clear how we build our UBI-based image, since we don't have a repo + CI for doing this currently.

nickboldt:
> It's not clear how we build our UBI-based image, since we don't have a repo + CI for doing this currently.

Yes, and that's a problem. How can you rely on an image if you have no idea how to build it? :( 

How about we use `quay.io/eclipse/che-quarkus:nightly` as the JDK 11 / maven image?

That's the one that's ref'd in https://github.com/eclipse/che-devfile-registry/pull/144 

I also don't know how it's built but i do see it as the last line in the list in https://github.com/eclipse/che-devfile-registry/blob/master/arbitrary-users-patch/base_images

svor:
> How about we use quay.io/eclipse/che-quarkus:nightly as the JDK 11 / maven image?

Right now it is quay.io/quarkus/centos-quarkus-maven:19.2.1 and it provides Java8, GraalVM, Maven, Podman and Buildah.

svor:
@amisevsk So, as we decided in the issue https://github.com/eclipse/che/issues/13926#issuecomment-559073072, I set empty value for MAVEN_CONFIG in devfiles where the sample uses **mvnw** utility.
If it's OK for now, i'll provide same fix for https://github.com/redhat-developer/devfile/tree/master/getting-started and https://github.com/redhat-developer/codeready-workspaces/tree/master/dependencies/che-devfile-registry/devfiles.

devfiles/apache-camel-springboot/devfile.yaml
```diff
@@ -26,8 +26,8 @@ components:
     alias: maven
     image: quay.io/eclipse/che-java8-maven:nightly
     env:
-      - name: MAVEN_CONFIG
-        value: "/home/user/.m2"
+      - name: MAVEN_CONFIG
```
amisevsk: Some stray whitespace here

devfiles/java-maven/devfile.yaml
```diff
@@ -18,8 +18,6 @@ components:
     alias: maven
     image: quay.io/eclipse/che-java11-maven:nightly
     env:
-      - name: MAVEN_CONFIG
-        value: /home/user/.m2
```
amisevsk: I think we should also unset `MAVEN_CONFIG` here, since the base image has `MAVEN_CONFIG=/root/.m2` by default. Our sample project doesn't use `mvnw` but it might trip up someone trying to import their own project.

svor: Unset for all maven based devfiles

PR #161, opened 2019-12-10
Update vert.x sample

svor:
### What does this PR do?
Updates vert.x sample

https://github.com/eclipse/che/issues/15314

### DIFF
--- a/devfiles/java-web-vertx/devfile.yaml
+++ b/devfiles/java-web-vertx/devfile.yaml
@@ -4,10 +4,10 @@ metadata:
   generateName: java-web-vertx-
 projects:
   -
-    name: java-web-vertx
+    name: vertx-http-example
     source:
       type: git
-      location: "https://github.com/che-samples/web-java-vertx"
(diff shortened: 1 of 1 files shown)


ibuziuk:
@svor  could you please also provide an update to the https://github.com/redhat-developer/devfile/tree/master/getting-started/vertx once this PR is merged ?

svor:
> @svor could you please also provide an update to the https://github.com/redhat-developer/devfile/tree/master/getting-started/vertx once this PR is merged ?

@ibuziuk looks like it already uses this sample, so don't need to change it.

ibuziuk:
@svor are the devfiles identical (other than `persistVolumes: 'false'` option) ?

svor:
@ibuziuk they have small differences, but they use the same sample and commands have same logic. Should they be identical?

ibuziuk:
@svor would be greate to make them identical (apart from the `persistVolumes: 'false'` option)

PR #162, opened 2019-12-10
Label arbitrary-uid-patched images with original base image

amisevsk:
### What does this PR do?
Add label 'che.base.image' to images built in arbitrary userID patch to store the original base image the patched image is build from.

It could be very useful to know which image the patched image was built off of, if e.g. we change base images at some point, or if a particular image causes issues for Che (e.g. the community maven image). As it stands it's hard (or maybe impossible) to figure out what the `$FROM_IMAGE` is for patched images.


amisevsk:
@nickboldt `podman` didn't like it, and the updated line seems to be the normal syntax; no idea where the `[--chown=0:0]` came from. The end result is the same as far as I can tell.

PR #164, opened 2019-12-12
Che4z-1.1.0

filipkroupa:
Added debugger-for-mainframe and hlasm-language-support

Signed-off-by: Filip Kroupa <filip.kroupa@broadcom.com>

### What does this PR do?

Update Che4z devfile to include newly added plugins [eclipse/che-plugin-registry#322](https://github.com/eclipse/che-plugin-registry/pull/322)

Targetting [Milestone 7.6.0](https://github.com/eclipse/che/milestone/107)

N/A

### DIFF
--- a/devfiles/che4z/devfile.yaml
+++ b/devfiles/che4z/devfile.yaml
@@ -17,10 +17,7 @@ components:
     id: broadcommfd/debugger-for-mainframe/latest
   - 
     type: chePlugin
-    id: broadcommfd/hlasm-language-support/latest    
-  - 
-    id: eclipse/che-theia/latest
-    type: cheEditor
+    id: broadcommfd/hlasm-language-support/latest
   - 
     alias: zowe-cli
     type: kubernetes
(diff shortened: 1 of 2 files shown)


devfiles/che4z/devfile.yaml
```diff
@@ -13,10 +13,13 @@ components:
...
   - 
-    id: eclipse/che-machine-exec-plugin/7.3.0
     type: chePlugin
+    id: broadcommfd/debugger-for-mainframe/latest
   - 
-    id: eclipse/che-theia/7.3.0
+    type: chePlugin
+    id: broadcommfd/hlasm-language-support/latest    
+  - 
+    id: eclipse/che-theia/latest
```
amisevsk: If speccing for 7.3.0 is no longer necessary, what do you think about removing a strict editor definition? By default, nightly builds of Che will use `eclipse/che-theia/latest`, and release builds of Che will use e.g. `eclipse/che-theia/7.6.0` (which would be `latest` for the `7.6.0` release), so including che-theia here shouldn't be required.

sleshchenko: +1 to Angel's proposal, it's good thing to do and I believe all devfiles in registry miss editor definition, if no - it should be fixed.

filipkroupa: I agree, since the version is no longer needed. I've checked on my test environment and removed the strict editor version definition.

PR #166, opened 2019-12-13
Add new devfile with sample based on yarn

svor:

### What does this PR do?
Adds a new devfile which contains sample based on yarn. The sample is React App generated by https://github.com/facebook/create-react-app and its location is https://github.com/che-samples/react-web-app.git. 

The provided commands are:

- install all required dependencies
- build the app
- run the app
- launch tests

https://github.com/eclipse/che/issues/14720

### DIFF
--- a/devfiles/nodejs-yarn/devfile.yaml
+++ b/devfiles/nodejs-yarn/devfile.yaml
@@ -1,0 +1,53 @@
+---
+apiVersion: 1.0.0
+metadata:
+  generateName: nodejs-web-app-
+projects:
+  -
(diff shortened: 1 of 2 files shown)


svor:
@amisevsk Thanks for trying it. 
On my local minishift instance I don't have such problems, but when i tried it on che.openshift.io I also see problems like you described.
I don't know what is the reason, will do investigation.

svor:
@amisevsk I've updated the sample and this PR, could you pls try it again if you have time. And be sure that EPHEMERAL MODE is active for the workspace

PR #173, opened 2019-12-18
Change memory required for Apache Camel

apupier:
as the new version 0.0.20 requires an higher value

Signed-off-by: Aurélien Pupier <apupier@redhat.com>

### DIFF
--- a/devfiles/apache-camel-springboot/devfile.yaml
+++ b/devfiles/apache-camel-springboot/devfile.yaml
@@ -16,11 +16,11 @@ components:
   -
     type: chePlugin
     id: redhat/vscode-apache-camel/latest
-    memoryLimit: 200Mi
+    memoryLimit: 260Mi
   -
     type: chePlugin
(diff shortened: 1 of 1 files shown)


apupier:
> Before merging we should ensure we can merge both this and [eclipse/che-plugin-registry#325](https://github.com/eclipse/che-plugin-registry/pull/325) atomically within this release window.

I think the devfile can be merged even without the plugin reguistry updated. So I would suggest to merge this PR first and when it has been rebuilt, to merge the other PR.
What do you think?

Can you precise the release window?

amisevsk:
@apupier My concern is that the 7.6.0 release is in progress, though I think the branch point for registries is already defined. @ibuziuk could confirm.

apupier:
Ilya answered in Slack that he believes I can merge to master safely.
I also noticed that there was already another merge to master since the version upgrade.

I'm merging :-)

PR #174, opened 2019-12-18
Move from stretch to alpine

gattytto:
Signed-off-by: Esteban Mañaricua <emanaricua@gmail.com>

### What does this PR do?
changes linux distro used for che-golang container

not reported

### DIFF
--- a/arbitrary-users-patch/base_images
+++ b/arbitrary-users-patch/base_images
@@ -1,6 +1,6 @@
 che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7
 che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
-che-golang-1.12         golang:1.12-stretch
+che-golang-1.12         golang:1.12-alpine
 che-java11-gradle       gradle:5.2.1-jdk11
 che-java11-maven        maven:3.6.0-jdk-11
 che-java8-maven         maven:3.6.1-jdk-8


gattytto:
I haven't created an issue yet

it is required for being able to unzip google protobuff compiler (protoc) which is used to generate code for various languages from .proto files. 

since the container has no root permissions, one can't just apt install unzip and continue in a script, sudo also is not included in the container so, since alpine version of the golang container includes unzip command by default, it solves the problem and allows for developers to create scripts and tasks that download the zip, put it in a known folder, add the folder to the path and start code generation tasks before building a project

gattytto:
problem is that buster version of the image does not include unzip command and it also throws an error when using gunzip to try unpack protoc

amisevsk:
If a specific dependency is not present in any of the official golang images, then I think the necessary solution following https://github.com/eclipse/che/issues/15476 would be to create a new image in the `che-dockerfiles` org that installs the required dependency.

nickboldt:
Or maybe move to a UBI8 based golang image that doesn't have the problem and has already been vetted as working for CRW 2. :)

gattytto:
1.13-buster image test:
root@153994df4ec3:/go# gunzip protoc-3.11.2-linux-x86_64.zip 
gzip: protoc-3.11.2-linux-x86_64.zip: unknown suffix -- ignored
root@153994df4ec3:/go# unzip protoc-3.11.2-linux-x86_64.zip 
bash: unzip: command not found
root@153994df4ec3:/go# zcat -d protoc-3.11.2-linux-x86_64.zip 
gzip: protoc-3.11.2-linux-x86_64.zip has more than one entry--rest ignored

amisevsk:
As a workaround for your use case, I suggest extending the che-golang-1.12 image:

```Dockerfile
FROM quay.io/eclipse/che-golang-1.12:nightly

USER 0
RUN apt-get update && \
    apt-get install -y zip && \
    apt-get clean all

# Download, unzip, and prepare workspace as needed.

USER 10001
```

gattytto:
> I don't think using the alpine version of this image is suitable for a development environment; the alpine version is meant to be used for building or running, and as such does not include `bash` and `git`, for example.

I agree, I should close this and create an issue that gets to a sane resolution path

gattytto:
https://github.com/eclipse/che/issues/15533

PR #180, opened 2020-01-07
provide script for doing devfile reg...

nickboldt:
provide script for doing devfile reg release, and producing a PR for master branch

Change-Id: Iea3a9fbbfab195da60edad42015352cb3fca5da9

### DIFF
--- a/RELEASE.sh
+++ b/RELEASE.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+# Release process automation script. 
+# Used to create branch/tag, update VERSION files and and trigger release by force pushing changes to the release branch 
+
+# set to 1 to actually trigger changes in the release branch
+TRIGGER_RELEASE=0 
+
+#REPO=git@github.com:eclipse/che-machine-exec
+#VERSION=7.7.0
+
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    '-t'|'--trigger-release') TRIGGER_RELEASE=1; shift 0;;
+    '-r'|'--repo') REPO="$2"; shift 1;;
+    '-v'|'--version') VERSION="$2"; shift 1;;
+  esac
+  shift 1
+done
+
+usage ()
+{
+  echo "Usage: $0 --repo [GIT REPO TO EDIT] --version [VERSION TO RELEASE] [--trigger-release]"
+  echo "Example: $0 --repo git@github.com:eclipse/che-machine-exec --version 7.7.0 --trigger-release"; echo
+}
+
+if [[ ! ${VERSION} ]] || [[ ! ${REPO} ]]; then
+  usage
+  exit 1
+fi
+
+# derive branch from version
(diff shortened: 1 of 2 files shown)


RELEASE.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+# Release process automation script. 
+# Used to create branch/tag, update VERSION files and and trigger release by force pushing changes to the release branch 
+
+# set to 1 to actually trigger changes in the release branch
+TRIGGER_RELEASE=0 
+
+#REPO=git@github.com:eclipse/che-machine-exec
+#VERSION=7.7.0
```
amisevsk: Leftover notes? I hope we're not planning on updating this to 7.8.0 in 3 weeks :)

RELEASE.sh
```diff
@@ -0,0 +1,97 @@
...
+if [[ ! ${VERSION} ]] || [[ ! ${REPO} ]]; then
+  usage
+  exit 1
+fi
+
+# derive branch from version
+BRANCH=${VERSION%.*}.x
+
+# work in tmp dir
+TMP=$(mktemp -d); mkdir -p "$TMP"; cd "$TMP" || exit 1
```
amisevsk: `mkdir -p "$TMP"` seems to be unnecessary here, since `mktemp` will create it.

Maybe also a `pushd` with a trap to `popd` on exit makes sense here?

RELEASE.sh
```diff
@@ -0,0 +1,97 @@
...
+git clone "${REPO}" -q
+cd "${REPO##*/}" || exit 1
+git fetch origin master:master
+git checkout master
+
+# create new branch off master + push to origin
+git branch "${BRANCH}" || git checkout "${BRANCH}" && git pull origin "${BRANCH}"
+git push origin "${BRANCH}"
+git fetch origin "${BRANCH}:${BRANCH}"
+git checkout "${BRANCH}"
```
amisevsk: This sequence isn't clear to me -- why do we need to checkout `${BRANCH}` multiple times?

RELEASE.sh
nickboldt: those be comments, yo. sample values. But yeah I can remove that

RELEASE.sh
nickboldt: fair point about mktemp... but no need to popd back since if the dir isn't created, we can't have pushd'd or cd'd into it.

RELEASE.sh
nickboldt: if creating the branch fails, then I figured I'd checkout and pull instead

RELEASE.sh
amisevsk: Fair enough

RELEASE.sh
amisevsk: Doesn't the script leave your shell in `/tmp` upon completion though?

nickboldt: no, it runs in its own subshell

PR #183, opened 2020-01-13
Add icons to repo and use them in devfiles

amisevsk:
### What does this PR do?
Adds icons gathered by @beaumorley (much thanks!) to the devfile registry, and specify them in devfile's `meta.yaml`s.

Currently, relative icon URLs are not resolved by the dashboard, so deploying these changes gives 404 icons. I've built a workaround image, `amisevsk/che-devfile-registry:icons`, which uses the airgap env var `CHE_DEVFILE_REGISTRY_URL` to serve the icons correctly.

### Screenshot
che-amisevsk-che-2 devtools-dev ext devshift net_dashboard_ (8)

### Still TODO:
- The Go icon had a very rectangular aspect ratio, which results in the edge being cut off:
    Screenshot from 2020-01-13 14-57-09
    For now, I've grabbed a new svg from the [go brand guide](https://blog.golang.org/go-brand), though it's still slightly cut off
- The images we're currently using are base64-encoded pngs embedded in svgs. Does it make more sense to use plain png files instead?
- Some icons can still be refined (perhaps as a separate issue)
    - Use the react logo for nodejs+react
    - Use the laravel logo for php+laravel
    - etc.

### Depends on
- [x] https://github.com/eclipse/che/pull/15642
- [x] Relative path support in dashboard https://github.com/eclipse/che/pull/15706

Part of https://github.com/eclipse/che/issues/15581

### Testing
Deploy `amisevsk/che-devfile-registry:icons` in place of the devfile registry, and set environment variable `CHE_DEVFILE_REGISTRY_URL` to the public route for the registry.


build/dockerfiles/Dockerfile
```diff
@@ -31,6 +31,7 @@ FROM registry.centos.org/centos/httpd-24-centos7 AS registry
 RUN mkdir -m 777 /var/www/html/devfiles
 COPY .htaccess README.md /var/www/html/
 COPY --from=builder /build/devfiles /var/www/html/devfiles
+COPY ./images /var/www/html/images
```
nickboldt: do we need to set 777 perms here too?

amisevsk: It shouldn't be necessary, since the files are by default group-readable and we're not looking to modify them. I suppose it could be done for safety, though.

amisevsk:
I've pushed a squashed commit with all of the changes. Currently, this PR:

- Uses [provided icons](https://github.com/eclipse/che/issues/15581#issuecomment-573256553) for devfiles (including updated `dotnetcore.svg`)
    - These had to be updated `xlink:href="data:img/png;base64 -> xlink:href="data:image/png;base64`, since for some reason Chrome does not render `data:img/png` and only handles `data:image/png`.
- Uses a different image for Go than is provided, as the aspect ratio resulted in the icon being cut off.
- Images are `png` files embedded in svg, with the exception of the Go icon.

How it looks now (with nightly Che server):
- Firefox:
Screenshot from 2020-01-22 13-16-42
Screenshot from 2020-01-22 13-16-50


- Chrome: 
che-amisevsk-che-2 devtools-dev ext devshift net_dashboard_
che-amisevsk-che-2 devtools-dev ext devshift net_dashboard_ (1)

devfiles/java-web-vertx/meta.yaml
```diff
@@ -2,5 +2,5 @@
 displayName: Java Vert.x
 description: Java stack with OpenJDK 8 and Vert.x demo application
 tags: ["Java", "OpenJDK", "Maven", "Vertx"]
-icon: https://www.eclipse.org/che/images/logo-eclipseche.svg
+icon: ./images/vertx.svg
```
sleshchenko: Since current directory `./` is used here, I expect that it should be resolved to `CHE_PLUGIN_REGISTRY_HOST`/devfiles/java-web-vertx/images/vertx.svg...
But in fact, it will be ``CHE_PLUGIN_REGISTRY_HOST`/images/vertx.svg...`, right?
If yes - can we use reference relative to root folder, like `/images/vertx.svg` ?

images/go.svg
```diff
@@ -0,0 +1 @@
... .6.3.6.7l-.2 2.4c0 .4-.4.7-.7.7z" id="path20" fill="#00acd7"/></g></g></g><g id="g39"><g id="CXHf1q_3_"><g id="g36"><g id="g30"><path class="st4" d="M153.1 99.3c-6.3 1.6-10.6 2.8-16.8 4.4-1.5.4-1.6.5-2.9-1-1.5-1.7-2.6-2.8-4.7-3.8-6.3-3.1-12.4-2.2-18.1 1.5-6.8 4.4-10.3 10.9-10.2 19 .1 8 5.6 14.6 13.5 15.7 6.8.9 12.5-1.5 17-6.6.9-1.1 1.7-2.3 2.7-3.7h-19.3c-2.1 0-2.6-1.3-1.9-3 1.3-3.1 3.7-8.3 5.1-10.9.3-.6 1-1.6 2.5-1.6h36.4c-.2 2.7-.2 5.4-.6 8.1-1.1 7.2-3.8 13.8-8.2 19.6-7.2 9.5-16.6 15.4-28.5 17-9.8 1.3-18.9-.6-26.9-6.6-7.4-5.6-11.6-13-12.7-22.2-1.3-10.9 1.9-20.7 8.5-29.3 7.1-9.3 16.5-15.2 28-17.3 9.4-1.7 18.4-.6 26.5 4.9 5.3 3.5 9.1 8.3 11.6 14.1.6.9.2 1.4-1 1.7z" id="path28" fill="#00acd7"/></g><g id="g34"><path class="st4" d="M186.2 154.6c-9.1-.2-17.4-2.8-24.4-8.8-5.9-5.1-9.6-11.6-10.8-19.3-1.8-11.3 1.3-21.3 8.1-30.2 7.3-9.6 16.1-14.6 28-16.7 10.2-1.8 19.8-.8 28.5 5.1 7.9 5.4 12.8 12.7 14.1 22.3 1.7 13.5-2.2 24.5-11.5 33.9-6.6 6.7-14.7 10.9-24 12.8-2.7.5-5.4.6-8 .9zm23.8-40.4c-.1-1.3-.1-2.3-.3-3.3-1.8-9.9-10.9-15.5-20.4-13.3-9.3 2.1-15.3 8-17.5 17.4-1.8 7.8 2 15.7 9.2 18.9 5.5 2.4 11 2.1 16.3-.6 7.9-4.1 12.2-10.5 12.7-19.1z" id="path32" fill="#00acd7"/></g></g></g></g></g></svg>
```
sleshchenko: new line it missing

beaumorley:
A couple of questions:

1.) Are the images OK from a tech/format perspective? I see above it was noted they "are png files embedded in svg, with the exception of the Go icon". Do you want svg? I have the masters and can send an updated .zip. 

2.) Do we need a logo for Maineframe?  I included this in the zip. But Mainframe is showing up with the Eclipse Che logo. Wondering if this was not the correct logo choice? https://github.com/openmainframeproject/artwork/tree/master/badges

Thanks!

devfiles/java-web-vertx/meta.yaml
amisevsk: I actually had `/images/vertx.svg` initially, but it looks like the dashboard is just prepending the devfile registry url -- viewing the requests in the console showed images being loaded from e.g.
```
http://che-devfile-registry.cluster.net//images/go.svg
```
That may be a dashboard issue though -- I'm not sure.

images/go.svg
amisevsk: Fixed.

amisevsk:
@beaumorley Good catch with the mainframe logo -- thanks! I had missed that one when updating the devfiles. I've updated the PR to use the correct logo.

As for `png` vs `svg`, I think it's not an issue currently for this PR -- I mostly wanted to note it in case someone wants to switch later or if we run into some issue with pngs (e.g. they might look strange if scaled). If you have svg icons handy, I can update this PR to use them, but I'm not worried about it in general.

beaumorley:
@amisevsk here are new logos. I redid them all, ensuring the aspect ratio issues were addressed as well as that they are .svg. When you have a chance, I suggest replacing all the ones you have with these. Thank you.

[final_logos_resized.zip](https://github.com/eclipse/che-devfile-registry/files/4105214/final_logos_resized.zip)

amisevsk:
thanks @beaumorley. New icons are pushed in the latest commit and available in docker image `amisevsk/che-devfile-registry:icons`

amisevsk:
One remaining issue with this PR is that it appears that resolving devfile icon relative paths doesn't work for the "Create workspace" page:

Screenshot from 2020-01-24 10-55-27

sleshchenko:
> One remaining issue with this PR is that it appears that resolving devfile icon relative paths doesn't work for the "Create workspace" page:

Is fixed in https://github.com/eclipse/che/pull/15832
Screenshot_20200127_125625

PR #185, opened 2020-01-16
update RELEASE.sh to support z-stream...

nickboldt:
update RELEASE.sh to support z-stream releases

Change-Id: If1d4f16eb12c1dd842a7db459e388a0d5afbe9fd


nickboldt:
tested this script by doing the 7.7.1 release, so I know it works

PR #188, opened 2020-01-28
Add docker image volume to mongo container in java-mongo devfile

amisevsk:
### What does this PR do?
The mongodb container in the java-mongo devfile mounts a volume, which causes it to be unrunnable on Hosted Che. 

This PR adds a volume for the mongodb container to make it runnable, but there's no way to force a volume such as this to be emptyDir, so we do introduce all the risks of storing more data in a PV (slow writes, potentially, etc.).

Fixes https://github.com/eclipse/che/issues/15850

### DIFF
--- a/devfiles/java-mongo/devfile.yaml
+++ b/devfiles/java-mongo/devfile.yaml
@@ -63,6 +63,9 @@ components:
       discoverable: 'true'
       public: 'false'
     port: 27017
+  volumes:
+  - name: mongodb_data
+    containerPath: /var/lib/mongodb/data
 commands:
 -
   name: maven build backend


sleshchenko:
> but there's no way to force a volume such as this to be emptyDir, so we do introduce all the risks of storing more data in a PV (slow writes, potentially, etc.).

In my understanding each volume should be transformed to emptyDir if a workspace is configured to use temporary storage(new wording for ephemeral mode). So, it would slow down only workspaces without temporary storage, right or do I miss something?

BTW I suppose it might be useful to configure specific volume to be temporary in addition to configuring it on Devfile level globally.

amisevsk:
> In my understanding each volume should be transformed to emptyDir if a workspace is configured to use temporary storage(new wording for ephemeral mode). So, it would slow down only workspaces without temporary storage, right or do I miss something?

Yeah, my main issue is that there's no reason to actually mount this as a persistent volume; there's no hard requirement that data in the testing db is persisted (and might make some things more confusing if there's a "configure database" command). Ideally, I'd like to see devfile volumes to be extended along the lines of
```yaml
  volumes:
  - name: mongodb_data
    containerPath: /var/lib/mongodb/data
    persistent: false
```
to force that specific volume to be created as emptyDir, regardless of the default. `persistent: true` would be default and would be overridden by Che server settings (if workspace is set as ephemeral overall).

PR #189, opened 2020-01-29
chore(github): Code Owners of devfile registry is members of Language…

benoitf:
### What does this PR do?

Based on last che SOS and based on repositories scope:

https://docs.google.com/document/d/1h6t4novpcRHjxe8DrEhTamDmit9VviOqcMQNtahM9uE/edit#heading=h.nh4qax6oermn

owners of che-devfile-registry is Thomas with languages team

Change-Id: I3e67500e8306a6b6ff848cdf892a86bd35a057fd

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # Global Owners
-* @tsmaeder @JPinkney @svor @sunix @benoitf
+* @tsmaeder @JPinkney @svor @sunix @benoitf @amisevsk @nickboldt
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 # Global Owners
-* @vparfonov @l0rd @rhopp @skabashnyuk @amisevsk @nickboldt @ibuziuk
+* @tsmaeder @JPinkney @svor @sunix @benoitf


nickboldt:
I would keep both @amisevsk & @nickboldt on the list as we do a lot of code review here too

benoitf:
why it has been modified before merging ?

nickboldt:
There was some disagreement about the value of removing me and Angel, and no one from the new Lang team objected so I fixed the PR and merged it. :)

Also until Eric is voted in as a committer, it makes sense to me to have a couple backup owners in his TZ who can push commits on his behalf.

If you want to instead fix the repo to allow non-codeowners to merge, then there's little point in having a codeowners file, in which case I'd have the same permissions you plan to bestow on my by making the repo config change.

TL;DR, is there a problem?

benoitf:
> Also until Eric is voted in as a committer, it makes sense to me to have a couple backup owners in his TZ who can push commits on his behalf.

Sorry but Eric is not owner of devfile registry.

> If you want to instead fix the repo to allow non-codeowners to merge, then there's little point in having a codeowners file, in which case I'd have the same permissions you plan to bestow on my by making the repo config change.

you can already merge without being code owners on this repository which is the default rule for all Eclipse Che repositories (except plugin registry but will be fixed)

PR #192, opened 2020-01-31
Fixes java-mysql tasks for latest version of petclinic

JPinkney:
Signed-off-by: Josh Pinkney <joshpinkney@gmail.com>

### What does this PR do?
This PR fixes an issue with the devfile's tasks that occured because of changes in the upstream version of the petclinic project. It also pins the java-mysql devfile to a specific commitId so that this won't happen again.

Fixes: https://github.com/eclipse/che/issues/15852
Reference: https://github.com/eclipse/che/issues/15898

### DIFF
--- a/devfiles/java-mysql/devfile.yaml
+++ b/devfiles/java-mysql/devfile.yaml
@@ -8,6 +8,7 @@ projects:
     source:
       type: git
       location: "https://github.com/spring-projects/spring-petclinic.git"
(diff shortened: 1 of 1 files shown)


nickboldt:
FWIW we have forks in use in CRW to avoid this problem and to have more control over the samples. 

https://github.com/che-samples

Could do the same for more Che samples as needed.

PR #195, opened 2020-02-13
Add option to use digests for images referenced in registry

amisevsk:
Note: this PR is a port of https://github.com/eclipse/che-plugin-registry/pull/379

### What does this PR do?
Add script /build/scripts/write_image_digests.sh, which will rewrite all image references in registry to use the current digest for each image tag. E.g.
```  
  - type: dockerimage
    alias: maven
    image: "quay.io/eclipse/che-java11-maven:nightly"
```
is replaced with
```
  - type: dockerimage
    alias: maven
    image: "quay.io/eclipse/che-java11-maven@sha256:1f8fb2175601f1b1cbb35a4250c817ada42dc6c388c710c696802cfba512f1ed" # tag: quay.io/eclipse/che-java11-maven:nightly
```
To enable this functionality, it is necessary to install `skopeo` in the builder images, which required changing adding the CentOS8-AppStream repo for the rhel build.

### Additional info
Digest-rewriting functionality is disabled by default and can be enabled by using the `./build.sh` option `--use-digests`, or by passing docker build arg `USE_DIGESTS=true`.

Existing airgap options are unaffected (you can still set env vars to override registry, organization, and tag). In the case of overriding tags, image digests are replaced with the tag specified.

There's also a commit included to update all copyright years for 2020.

### Testing
Image is available as `amisevsk/che-devfile-registry:digests`. I've tested the changes on the dev cluster and locally.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -14,6 +14,7 @@ REGISTRY="quay.io"
 ORGANIZATION="eclipse"
 TAG="nightly"
 TARGET="registry"
+USE_DIGESTS=false
 DOCKERFILE="./build/dockerfiles/Dockerfile"
 
 USAGE="
@@ -27,6 +28,8 @@ Options:
         Docker registry to be used for image; default 'quay.io'
     --organization, -o [ORGANIZATION]
         Docker image organization to be used for image; default: 'eclipse'
+    --use-digests
+        Build registry to use images pinned by digest instead of tag
     --offline
         Build offline version of registry, with all sample projects
         cached in the registry; disabled by default.
@@ -54,6 +57,10 @@ function parse_arguments() {
             ORGANIZATION="$2"
             shift; shift;
             ;;
+            --use-digests)
+            USE_DIGESTS=true
+            shift
+            ;;
             --offline)
             TARGET="offline-registry"
             shift
@@ -76,10 +83,19 @@ VERSION=$(head -n 1 VERSION)
 case $VERSION in
   *SNAPSHOT)
     echo "Snapshot version (${VERSION}) specified in $(find . -name VERSION): building nightly plugin registry."
-    docker build -t "${IMAGE}" -f ${DOCKERFILE} --target ${TARGET} .
+    docker build \
+        -t "${IMAGE}" \
+        -f ${DOCKERFILE} \
+        --build-arg "USE_DIGESTS=${USE_DIGESTS}" \
+        --target ${TARGET} .
     ;;
   *)
     echo "Release version specified in $(find . -name VERSION): Building plugin registry for release ${VERSION}."
-    docker build -t "${IMAGE}" -f ${DOCKERFILE} --target ${TARGET} --build-arg "PATCHED_IMAGES_TAG=${VERSION}" .
+    docker build \
+        -t "${IMAGE}" \
+        -f ${DOCKERFILE} \
+        --build-arg "USE_DIGESTS=${USE_DIGESTS}" \
+        --build-arg "PATCHED_IMAGES_TAG=${VERSION}" \
+        --target ${TARGET} .
     ;;
 esac
--- a/build/dockerfiles/content_sets_centos8_appstream.repo
+++ b/build/dockerfiles/content_sets_centos8_appstream.repo
@@ -1,0 +1,5 @@
+[AppStream]
+name=CentOS-8 - AppStream
+baseurl=http://mirror.centos.org/centos-8/8/AppStream/x86_64/os/
+gpgcheck=0
+enabled=1
--- a/build/dockerfiles/rhel.Dockerfile
+++ b/build/dockerfiles/rhel.Dockerfile
@@ -21,8 +21,8 @@ USER 0
 
 ARG BOOTSTRAP=false
 ENV BOOTSTRAP=${BOOTSTRAP}
-ARG LATEST_ONLY=false
-ENV LATEST_ONLY=${LATEST_ONLY}
+ARG USE_DIGESTS=false
+ENV USE_DIGESTS=${USE_DIGESTS}
 
 # to get all the python deps pre-fetched so we can build in Brew:
 # 1. extract files in the container to your local filesystem
@@ -39,7 +39,7 @@ ENV LATEST_ONLY=${LATEST_ONLY}
 
 # NOTE: uncomment for local build. Must also set full registry path in FROM to registry.redhat.io or registry.access.redhat.com
 # enable rhel 7 or 8 content sets (from Brew) to resolve jq as rpm
-COPY ./build/dockerfiles/content_sets_epel7.repo /etc/yum.repos.d/
+COPY ./build/dockerfiles/content_sets_centos8_appstream.repo /etc/yum.repos.d/
 
 COPY ./build/dockerfiles/rhel.install.sh /tmp
 RUN /tmp/rhel.install.sh && rm -f /tmp/rhel.install.sh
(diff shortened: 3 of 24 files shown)


build.sh
```diff
@@ -76,10 +83,19 @@ VERSION=$(head -n 1 VERSION)
 case $VERSION in
   *SNAPSHOT)
     echo "Snapshot version (${VERSION}) specified in $(find . -name VERSION): building nightly plugin registry."
-    docker build -t "${IMAGE}" -f ${DOCKERFILE} --target ${TARGET} .
+    docker build \
+        -t "${IMAGE}" \
+        -f ${DOCKERFILE} \
+        --build-arg "USE_DIGESTS=${USE_DIGESTS}" \
```
nickboldt: does not work for Brew builds. Need another approach, like using an ENV var or multistage.

build/dockerfiles/content_sets_centos8_appstream.repo
```diff
@@ -0,0 +1,5 @@
+[AppStream]
```
nickboldt: ya'll should just switch to UBI and reuse what I already have downstream for this.

build/dockerfiles/rhel.Dockerfile
```diff
@@ -21,8 +21,8 @@ USER 0
 
 ARG BOOTSTRAP=false
 ENV BOOTSTRAP=${BOOTSTRAP}
-ARG LATEST_ONLY=false
-ENV LATEST_ONLY=${LATEST_ONLY}
+ARG USE_DIGESTS=false
+ENV USE_DIGESTS=${USE_DIGESTS}
```
nickboldt: ok, I can probably use this in Brew if I overwrite with `env USE_DIGESTS=true`

build/dockerfiles/rhel.Dockerfile
```diff
@@ -39,7 +39,7 @@ ENV LATEST_ONLY=${LATEST_ONLY}
 
 # NOTE: uncomment for local build. Must also set full registry path in FROM to registry.redhat.io or registry.access.redhat.com
 # enable rhel 7 or 8 content sets (from Brew) to resolve jq as rpm
-COPY ./build/dockerfiles/content_sets_epel7.repo /etc/yum.repos.d/
+COPY ./build/dockerfiles/content_sets_centos8_appstream.repo /etc/yum.repos.d/
```
nickboldt: DUDE. You got Centos in my RHEL. Not cool, bro.

build/scripts/write_image_digests.sh
```diff
@@ -0,0 +1,21 @@
...
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+
+readarray -d '' devfiles < <(find "$1" -name 'devfile.yaml' -print0)
+for image in $(yq -r '.components[]?.image' "${devfiles[@]}" | grep -v "null" | sort | uniq); do
+  echo "Rewriting image $image"
+  digest=$(skopeo inspect "docker://${image}" | jq -r '.Digest')
```
nickboldt: This works with one caveat: the sha256sum for the image pushed to Quay is not the same as the image pushed to RHCC.

amisevsk: Wouldn't the build take place with the images set as desired? I didn't realize we were building the registry and then updating the images inside it.

davidfestal: I assume you would run this build script for both the registry image destinated to quay, and the registry image destinated to rhcc anyway. So it should not be a problem.

build/dockerfiles/content_sets_centos8_appstream.repo
amisevsk: It [is UBI](https://github.com/eclipse/che-devfile-registry/blob/master/build/dockerfiles/rhel.Dockerfile#L15), but none of the UBI repos contain skopeo AFAICT.

build.sh
amisevsk: I don't know that it's possible to pass an env var into the docker context, `build-arg` seems to be the only option for setting build-time variables.

build/dockerfiles/rhel.Dockerfile
amisevsk: Give me a RHEL repo that contains skopeo and doesn't require a subscription and I'll toss it in here :)

build/dockerfiles/rhel.Dockerfile
amisevsk: I figured there was a reason to have `ENV BOOTSTRAP` above. I can add this to the plugin registry PR as well.

build/scripts/write_image_digests.sh
nickboldt: > and then updating the images inside it.

So this change means that if we have to rebuild any of the stacks-* or plugin-* containers used in the plugin and devfile registries, we have to ALSO then rebuild the registries to pull in the new SHAs. Yay for extra churn. :)

> run this build script for both 

yes, [CRW-214](https://issues.redhat.com/browse/CRW-214) will solve all our problems here :D

build/dockerfiles/rhel.Dockerfile
nickboldt: well, I need to find one for this to work downstream because Brew can't see epel and centos repos so ... challenge accepted & required.

build/dockerfiles/rhel.Dockerfile
nickboldt: Sweet, please do.

build/dockerfiles/content_sets_centos8_appstream.repo
nickboldt: well, the Brew internal ones do, such as **REDACTED**/content/dist/rhel8/8/x86_64/appstream/os/Packages/s/

build.sh
nickboldt: multistage would work here too, but be clunky for upstream.

build/dockerfiles/rhel.Dockerfile
nickboldt: skopeo is in the appstream repo for RHEL8 so to build this locally we have three options:
* RHEL subscription
* use Brew internal pulp .repo
* use centos public .repo
 
Though I really thought that rpms in UBI were also free w/o subscription. Hmm. Maybe not?

build/dockerfiles/content_sets_centos8_appstream.repo
amisevsk: Sounds like *someone* is going to be using a custom `.repo` in the Brew build :wink:

nickboldt: Nope, in Brew we don't need no steenkin' repos. Brew handles that for us. (Because it's locked down to only approved repos & rpms.)

l0rd:
For the records I had some concerns that I have shared with @amisevsk yesterday. Mainly I was afraid that going in this direction (calculate the digest at build time) is against the goal to replace tags with digests for the images referenced in this repo. 

Replacing tags with digests in the git repo would have the merit to:
- make the build of the registries repeatable (with [that as well](https://github.com/eclipse/che/issues/15898))
- reduce the number of replacements in the devfiles when preparing a release
- reducing the number of builds and push of images (no changes --> no build/push)

But discussing with @amisevsk we agreed that may more complicated then this PR and we could do the replacement of the tags with digests during next sprint.

l0rd:
Just to be clear I am ok to merge this PR and the one on the plugin-registry side as they are. Just saying that next sprint we should [start working on using digests only](https://github.com/eclipse/che/issues/16047).

davidfestal:
> Just to be clear I am ok to merge this PR and the one on the plugin-registry side as they are. Just saying that next sprint we should [start working on using digests only](https://github.com/eclipse/che/issues/16047).

Thanks @l0rd !

These PRs should be very useful until we can generalize the use of digests.

PR #198, opened 2020-02-24
Increase memory for java-mysql petclinic devfile

JPinkney:
Signed-off-by: Josh Pinkney <joshpinkney@gmail.com>

### What does this PR do?
This PR increases the memory for java-mysql devfile so that it won't run into memory issues when running `maven build`. Tested locally and on che.openshift.io.

https://github.com/eclipse/che/issues/15911


JPinkney:
@svor Good catch! I've updated the meta.yaml as well

PR #200, opened 2020-02-24
push downstream fixes for CRW back into...

nickboldt:
push downstream fixes for CRW back into upstream where applicable

Change-Id: I3470eec087de4f8bf0a04eefc77b4a84d989b80e

### DIFF
--- a/build/dockerfiles/rhel.cache_projects.sh
+++ b/build/dockerfiles/rhel.cache_projects.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+#
+# Copyright (c) 2018-2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+set -x
+
+if [[ ! -f /tmp/resources.tgz ]] || [[ ${BOOTSTRAP} == "true" ]]; then
+  ./cache_projects.sh devfiles resources
+else
+  # unpack into specified folder
+  tar -xvf /tmp/resources.tgz -C "$1/"
+fi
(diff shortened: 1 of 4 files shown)


build/dockerfiles/rhel.install.sh
```diff
@@ -11,7 +11,10 @@
 microdnf install -y findutils bash wget yum gzip git tar jq python3-six python3-pip skopeo && microdnf -y clean all && \
 # install yq (depends on jq and pyyaml - if jq and pyyaml not already installed, this will try to compile it)
 if [[ -f /tmp/root-local.tgz ]] || [[ ${BOOTSTRAP} == "true" ]]; then \
-    mkdir -p /root/.local; tar xf /tmp/root-local.tgz -C /root/.local/; rm -fr /tmp/root-local.tgz;  \
+    mkdir -p /root/.local; \
+	if [[ -f /tmp/root-local.tgz ]]; then \
+		tar xf /tmp/root-local.tgz -C /root/.local/; rm -fr /tmp/root-local.tgz; \
+	fi; \
```
amisevsk: Side question: Why are the lines in this script merged?

build/dockerfiles/rhel.cache_projects.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
amisevsk: I'm not sure this script belongs in this repo -- it's not referenced by the default rhel.Dockerfile and is *only* used in the downstream build.

nickboldt: but what does it hurt, being upstream and colocated with rhel.Dockerfile?

build/dockerfiles/rhel.install.sh
nickboldt: ... because apparently I got into that habit w/ dockerfiles? *shrug*

build/dockerfiles/rhel.cache_projects.sh
amisevsk: It's strange to include a script that's never used, but is required for downstream -- someone could modify this script (or remove it, after noticing that it's unused) and break downstream.

nickboldt: ok, sure but the rhel.Dockerfile is also never used in upstream but required in down. 

Including it in up serves the purpose of reminding people to think about downstream when making changes in up.

amisevsk: Fair enough -- it's your call. However, rhel.Dockerfile *is* used, in the sense that it's a build option in this repository; the script is not standalone and is called nowhere. It's not even clear in the context of this repo where it should be called (since in midstream we call it during the builder phase instead of the `offline-builder` phase).

PR #205, opened 2020-03-04
Update gradle image to use Gradle6; Add user to /etc/group

svor:

### What does this PR do?
Updates **che-java11-gradle** image to use Gradle v.6.2.1;
Patches `/etc/group` by adding arbitrary user

@amisevsk @nickboldt I added current arbitrary user into /etc/group because I got a message after opening the terminal:
` groups: cannot find name for group ID 1000160000`
Any concerns of that?

https://github.com/eclipse/che/issues/15723

### DIFF
--- a/arbitrary-users-patch/base_images
+++ b/arbitrary-users-patch/base_images
@@ -1,7 +1,7 @@
 che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7
 che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
 che-golang-1.12         golang:1.12-stretch
-che-java11-gradle       gradle:5.2.1-jdk11
+che-java11-gradle       gradle:6.2.1-jdk11
 che-java11-maven        maven:3.6.0-jdk-11
 che-java8-maven         maven:3.6.1-jdk-8
 che-nodejs10-community  node:10.16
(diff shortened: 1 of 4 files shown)


svor:
@amisevsk you can try this devfile https://gist.githubusercontent.com/svor/0fe771b1f1e853a98ee38442b75023bd/raw/f9c994c4316927cfb1fc57b37eb9beaf069a4dba/GradleDevfile.

tsmaeder:
@svor I think we need to understand how groups are used before doing the groups change. If the error message (group has no name) existed before, I propose we reduce the PR to just the version update.

svor:
@tsmaeder the error message didn't exist before i can reproduce this message only with new gradle image (https://hub.docker.com/r/amd64/gradle)

arbitrary-users-patch/base_images
```diff
@@ -1,7 +1,7 @@
 che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7
 che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
 che-golang-1.12         golang:1.12-stretch
-che-java11-gradle       gradle:5.2.1-jdk11
+che-java11-gradle       gradle:6.2.1-jdk11
```
amisevsk: The corresponding meta.yaml for gradle should be updated, since it still refers to 5.2.1 in description.

amisevsk:
Looking into the groups issue some more, I have no idea what's going on. On the gradle 5.2.1 image, I have:
```bash
user@workspacefjanf44yq6wpgnro:/projects$ id
uid=1000420000(user) gid=0(root) groups=0(root),1000420000
```
The group has no name, but the message is not printed.

On 6.2.1:
```bash
user@workspacehzwccyj8cswa4iqh:/projects$ id
uid=1000420000(user) gid=0(root) groups=0(root),1000420000
```
and the error message. :man_shrugging:

arbitrary-users-patch/base_images
svor: fixed

PR #206, opened 2020-03-05
fix support for newer git (doesn't work with...

nickboldt:
fix support for newer git (doesn't work with --remote=path/on/local; use pushd and simpler git archive command so we can test this script locally);
fix support for dotnet31 devfile, which doesn't use master branch;
tweak console output and arg desc (cache_project REQUIRES branch, it's not optional)

Change-Id: Ia7ffb1791b04e83456be9e1a978b69ee1116fb36

### DIFF
--- a/build/scripts/cache_projects.sh
+++ b/build/scripts/cache_projects.sh
@@ -24,14 +24,17 @@ TEMP_REPO="${TEMP_DIR}/cloned"
 # Clone a git repository and create an archive zip at a specified location
 # Args:
 #   $1 - URL of git repo
-#   $2 - (optional) branch to archive
+#   $2 - branch to archive
 #   $3 - destination path for the archived project zip file
 function cache_project() {
   local repo="$1"
   local branch="$2"
   local destination="$3"
-  git clone "$repo" -b "$branch" --depth 1 "$TEMP_REPO" &>/dev/null
-  git archive "$branch" --remote="$TEMP_REPO" --format zip --output "$destination"
+  rm -fr "$TEMP_REPO"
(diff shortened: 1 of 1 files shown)


nickboldt:
See more discussion/context/explanation in https://github.com/redhat-developer/codeready-workspaces/pull/247 re: https://issues.redhat.com/browse/CRW-644

nickboldt:
when approved, should also cherry pick to 7.9.x

build/scripts/cache_projects.sh
```diff
@@ -24,14 +24,17 @@ TEMP_REPO="${TEMP_DIR}/cloned"
...
 function cache_project() {
   local repo="$1"
   local branch="$2"
   local destination="$3"
-  git clone "$repo" -b "$branch" --depth 1 "$TEMP_REPO" &>/dev/null
-  git archive "$branch" --remote="$TEMP_REPO" --format zip --output "$destination"
+  rm -fr "$TEMP_REPO"
+  git clone "$repo" -b "$branch" --depth 1 "$TEMP_REPO" -q
+  pushd "$TEMP_REPO" &>/dev/null
+    git archive "$branch" -o "$destination"
```
JPinkney: Should this command be in line with the other ones?

nickboldt: Could be, but that block is "pushd" in. ;)

(Personal style choice. Helps me remember when I'm in a pushd/popd block.)

JPinkney: Ah gotcha!, that makes sense!

PR #212, opened 2020-03-16
Support sparseCheckoutDir for offline cached projects

amisevsk:
### What does this PR do?
Support the `sparseCheckoutDir` option when caching offline projects. Turns out it's not hard to do (though it took some time before I noticed the option in `git-archive`...)

Camel-k stack has additional commands present in Theia due to package.json files elsewhere in the repo.

### DIFF
--- a/build/scripts/cache_projects.sh
+++ b/build/scripts/cache_projects.sh
@@ -26,14 +26,22 @@ TEMP_REPO="${TEMP_DIR}/cloned"
 #   $1 - URL of git repo
 #   $2 - branch to archive
 #   $3 - destination path for the archived project zip file
+#   $4 - sparse checkout directory
 function cache_project() {
   local repo="$1"
   local branch="$2"
   local destination="$3"
+  local sparse_checkout_dir="$4"
+
   rm -fr "$TEMP_REPO"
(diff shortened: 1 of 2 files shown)


nickboldt:
This simple fix seems quite reasonable... I'm jut not sure if the zip having all its contents in a subfolder will work when unpacking into the offline devfile.

Compare:

image

But this should at least solve the problem in https://issues.redhat.com/browse/CRW-695, so +1.

nickboldt:
But at least for the quarkus devfile, `getting-started` folder IS configured correctly..

image

```
commands:
  - name: Package the application
    actions:
      - type: exec
        component: maven
        command: mvn package
        workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
  - name: Start development server (debug mode)
    actions:
      - type: exec
        component: maven
        command: mvn compile quarkus:dev
        workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
```

amisevsk:
It duplicates the behavior you get when using sparse checkout in non-offline scenarios -- you get the same subfolder.

PR #215, opened 2020-03-23
Use golang tool with non go modules mod to fix intelligence

AndrienkoAleksandr:
### What does this PR do?
Use golang tool with non go modules to fix file imports intelligence, because we are using sample without go modules.

https://github.com/eclipse/che/issues/16427


ericwill:
I tested the sample Go workspace and all looks well. @amisevsk can you take a look?

ericwill:
> LGTM. I did encounter the known issue of analysis tools missing, but don't believe that's directly related to this PR.

It's not, I'm about to fix that in a separate PR.

PR #217, opened 2020-03-25
Update PHP base image

ericwill:
Update base image for PHP to that which is located in che-dockerfiles/che-php-base.

See eclipse/che#15854


ericwill:
Waiting until the base image is merged and built in che-dockerfiles.

PR #220, opened 2020-03-31
Use 20.0.0-java8 tag of centos-quarkus-maven

sunix:
Use latest tag of centos-quarkus-maven: 20.0.0-java8
Needed to have native compilation working with the default Quarkus sample

### DIFF
--- a/arbitrary-users-patch/base_images
+++ b/arbitrary-users-patch/base_images
@@ -10,5 +10,5 @@ che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
 che-php-7               quay.io/eclipse/che-php-base:7.4
 che-python-3.6          centos/python-36-centos7:1
 che-python-3.7          python:3.7.4-slim
-che-quarkus             quay.io/quarkus/centos-quarkus-maven:19.2.1
+che-quarkus             quay.io/quarkus/centos-quarkus-maven:20.0.0-java8
 che-rust-1.39           rust:1.39.0-slim
--- a/devfiles/quarkus/meta.yaml
+++ b/devfiles/quarkus/meta.yaml
@@ -1,6 +1,6 @@
 ---
 displayName: Quarkus Tools
-description: Quarkus Tools with OpenJDK 8 and Maven 3.6.0
+description: Quarkus Tools with OpenJDK 8 and Maven 3.6.3
 tags: ["Java", "Quarkus", "OpenJDK", "Maven", "Debian"]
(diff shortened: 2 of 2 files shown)


arbitrary-users-patch/base_images
```diff
@@ -10,5 +10,5 @@ che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
 che-php-7               quay.io/eclipse/che-php-base:7.4
 che-python-3.6          centos/python-36-centos7:1
 che-python-3.7          python:3.7.4-slim
-che-quarkus             quay.io/quarkus/centos-quarkus-maven:19.2.1
+che-quarkus             quay.io/quarkus/centos-quarkus-maven:20.0.0-java8
```
benoitf, line 13: maybe we could bump to Java11 version as well ? (before it was only providing java8)

sunix:
@svor  have you tested the patched image ? or ? could you provide a devfile ?

sunix:
@svor we have the same issue with the default stack

sunix:
looks like this is an issue with 1.3.1-Final, https://github.com/quarkusio/quarkus-quickstarts/commit/ceb711a04d59c5f8a7bbf8e1e7731c23e2d93acd
shouldn't we use a tagged version ? unless we have that stack tested in our CI.

svor:
@sunix I think we can do that or reference to some specific commit. Could you please update the devfile and try if it works. Or should we crate another issue for that?

sunix:
@svor it's going to be fixed in the next days I hope https://github.com/quarkusio/quarkus-images/issues/63 so let's wait and I will update my PR with the new version.

sunix:
@svor https://github.com/quarkusio/quarkus-images/issues/63#issuecomment-609911773
Are you sure you have tested with 20.0.0-java8 ?

svor:
@sunix I double check it and can confirm that it works nice with quay.io/quarkus/centos-quarkus-maven:20.0.0-java8 :thumbsup:
Just one remark we have to change [meta.yaml](https://github.com/eclipse/che-devfile-registry/blob/master/devfiles/quarkus/meta.yaml#L3) to point new mvn version

sunix:
@svor could you review again ?
also, this one should be backported to CRW and included in the next Che release as a blocker WDYT ?

arbitrary-users-patch/base_images
sunix: Do you want to use Java11 instead of 8 ? or provide both alternatives in our base image ?

PR #224, opened 2020-04-09
Fix memoryLimit to Mi instead of Mib since the latest is not valid anymore

mingfang:
### What does this PR do?
Fix memoryLimit to Mi instead of Mib since the latest is not valid anymore

https://github.com/eclipse/che/issues/16572

### DIFF
--- a/devfiles/java-mongo/devfile.yaml
+++ b/devfiles/java-mongo/devfile.yaml
@@ -12,7 +12,7 @@ components:
 -
   type: chePlugin
   id: redhat/java/latest
-  memoryLimit: 1280MiB
+  memoryLimit: 1280Mi
 -
   type: dockerimage
   alias: maven
--- a/devfiles/java-mysql/devfile.yaml
+++ b/devfiles/java-mysql/devfile.yaml
@@ -13,7 +13,7 @@ components:
   -
     type: chePlugin
     id: redhat/java/latest
-    memoryLimit: 1280MiB
+    memoryLimit: 1280Mi
   -
     type: dockerimage
     alias: tools


ibuziuk:
@mingfang hmm but  `memoryLimit: 1280MiB` was set from the very beginning and it used to work just fine. I suspect there is smth. more subtle

mingfang:
> @mingfang hmm but `memoryLimit: 1280MiB` was set from the very beginning and it used to work just fine. I suspect there is smth. more subtle

According to the spec, https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-requests-and-limits-of-pod-and-container, `MiB` is not valid.

Fabric8 added additional checks to follow the spec.
https://github.com/fabric8io/kubernetes-client/commit/c370d2628aa5da6b7217486373e16a35d7013df2#diff-218cae2eb257d95a3b82ec8d08ee17ed

JPinkney:
@mingfang In order for us to merge this PR you will to sign the eclipse eca and sign off your commits. Also do you mind updating the description with the relevant details!

mingfang:
> @mingfang will you be able to provide the same hot-fix for `7.11.x` branch ?
> Verified on Hosted Che and it works like a charm. Still not clear how it used to work before - those limits were in place since the very beggining

We might have upgraded the fabric8 version only recently.

ibuziuk:
sent a PR for 7.11.x since we need to fix Hosted Che - https://github.com/eclipse/che-devfile-registry/pull/225
reviews are most welcome

mingfang:
> @mingfang In order for us to merge this PR you will to sign the eclipse eca and sign off your commits. Also do you mind updating the description with the relevant details!

I tried the git amend and rebase but my git-fu is just not good enough to do it.
FWIW here is my signoff `Signed-off-by: Ming Fang <mingfang@mac.com>`

PR #226, opened 2020-04-17
feat(alpine): Change parent image to be alpine based as well

benoitf:
### What does this PR do?

It replaces centos/httpd with alpine/httpd allowing to be smaller/updated and also multi-arch

Change-Id: I3b9802799fcd05bb388b4fcd3ffa14a6424a8a9e

### DIFF
--- a/build/dockerfiles/Dockerfile
+++ b/build/dockerfiles/Dockerfile
@@ -6,7 +6,7 @@
 #
 # SPDX-License-Identifier: EPL-2.0
 #
-FROM alpine:3.10 AS builder
+FROM docker.io/alpine:3.11.5 AS builder
 RUN apk add --no-cache py-pip jq bash wget git skopeo && pip install yq
 
 # Registry, organization, and tag to use for base images in dockerfiles. Devfiles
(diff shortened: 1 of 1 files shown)


nickboldt:
Kinda wish we used UBI here so that we didn't have to maintain two radically different dockerfiles but *shrug-emoji* at least we have rhel.Dockerfile to sync with downstream.

Once the registries switch to being OOTB offline mode (in order to work with OCP 4.3+ & include digests) ... then the size difference between alpine and UBI might be negligible.

WDYT?

benoitf:
offline mode is not the default so size still count

nickboldt:
> offline mode is not the default so size still count

Yes, agreed TODAY. But what happens when we move Che to including digests? 

https://github.com/eclipse/che/issues/15350

benoitf:
I don't see what it's changing except replacing tags by the digests in meta.yaml ?

benoitf:
@ericwill @rhopp is it ok for you ?

rhopp:
Fine with me.

PR #229, opened 2020-04-24
dev mode + ./mvnw for quarkus

maxandersen:
Why:

 * "Development server" is not the right term for quarkus:dev (since 1.4
   there command mode which is not a server.
 * All Quarkus apps and examples by default comes with mvn wrapper. That
   should be the default used.

This change addreses the need by:

 * Make 'development server' 'Development mode"
 * Update tasks to use `./mvnw` instead of `mvn`

https://github.com/eclipse/che/issues/16701

### DIFF
--- a/devfiles/quarkus/devfile.yaml
+++ b/devfiles/quarkus/devfile.yaml
@@ -44,15 +44,15 @@ commands:
       -
         type: exec
         component: maven
-        command: "mvn package"
+        command: "./mvnw package"
         workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
   -
-    name: Start development server
+    name: Start Development mode
     actions:
       -
         type: exec
         component: maven
-        command: "mvn compile quarkus:dev"
+        command: "./mvnw compile quarkus:dev"
         workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
   -
     name: Attach remote debugger


ericwill:
LGTM, @maxandersen the ECA check failed -- please sign your commit and force push the branch it should clear it.

maxandersen:
> LGTM, @maxandersen the ECA check failed -- please sign your commit and force push the branch it should clear it.

done.

maxandersen:
> Hello, why do we have mvnw wrapper ?
> As we provide the image we don't need to download something with the wrapper

why we have the wrapper ? because then it is controlled what is used. In this case the mvn in the image is too old. in other cases it could also be too new. mvnw and gradle wrappers is what we would recommend users to use since then your build setup is not affected by your local or in this case (che) remote setup.

benoitf:
The issue is that you're post-downloading stuff which is not the promise of the container where you've everything setup for you.
So each time you run the action, it's downloading maven.

benoitf:
Also it should never be too old or too new as the project being checkout should be pinned to a specific sha1.

maxandersen:
@benoitf these are all issues on how che operates IMO. che should be able to precook the volume using the ./mvnw and gradle wrappers  instead of insisting users need to not only maintain a docker image *only* for che and then yet another setup for other IDE's.

I don't follow how a devfile will ever be valuable if the project it points to can't be a branch.

ericwill:
FWIW we have some cases where we do something similar, for example in the Go stack the tooling is installed based on the latest available Go package.

maxandersen:
@ericwill yes, exactly. it is not at all uncommon and it lets all workflows in and out of Che stay stable/similar and managed by the projects git state rather than external devfile mechanics.

benoitf:
It's not uncommon but the stacks inside Che should work in a predefined state.
Let say quarkus example update to Java11 ( which will happen) and user is using an old version of Che, then the experience is that it will fail as Java8 will be used.

This is why I'm saying that is provided by che is like a 'freeze' of something working

That said, if there is a devfile inside the quarkus repository, then there is no issue to use mvnw or other stuff because it's not 'provided/delivered' by che.

maxandersen:
@benoitf che's freeze is too global imo. 

There have and probably won't be a time where it ever makes sense to use an external mvn or gradle with most new java apps. It just doesn't help. For it to work the che team will need to provide NxM combinations of quarkus devfile.

Much more sustainable we can find a way to dynamically adapt - and then when users actually want to lock things down *then* they can put a devfile into their project. Until then its just going to cause a lot of unnecessary churn.

in other words - what is the value of che.openshift.io to have a *single* locked down devfile for a fast moving stack ? it would never really work - not even with the latest unless you constantly rebuild/reconfigures that stack.

PR #231, opened 2020-04-27
Quick fix all endpoints name

sunix:
Fix for https://github.com/eclipse/che/issues/16748

root bug: https://github.com/eclipse/che/issues/16381


vitaliy-guliy:
It looks like it's internal bug, the changing of the names is a workaround.

sunix:
@vitaliy-guliy the real bug is https://github.com/eclipse/che/issues/16381

PR #232, opened 2020-04-27
Change tags on digests

vitaliy-guliy:

### What does this PR do?
Use digests instead of tags when referencing docker images in devfiles.

Solves https://github.com/eclipse/che/issues/16047

### DIFF
--- a/devfiles/angular/devfile.yaml
+++ b/devfiles/angular/devfile.yaml
@@ -15,7 +15,8 @@ components:
   -
     type: dockerimage
     alias: nodejs
-    image: quay.io/eclipse/che-nodejs10-community:nightly
+    # image: quay.io/eclipse/che-nodejs10-community:nightly
+    image: quay.io/eclipse/che-nodejs10-community@sha256:67e8f270c1cadb1cee0052554145aceefbafe67a127115ed9e5573b0f6e4f4f4
     memoryLimit: 1Gi
     endpoints:
       - name: 'angular'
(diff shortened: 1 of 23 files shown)


vitaliy-guliy:
Tags on digests have been changed by hands.
I used to run `build.sh` in the repository root to check the build.
Though the original issue https://github.com/eclipse/che/issues/16047 describes only changing the references, from your feedback I see that we have to do something else.

nickboldt:
So... your vision is that we would run build.sh manually every time a new nightly image referred in the registry is released? 

How is that manageable? Surely you would prefer if the digests were updated automatically?

ericwill:
AFIUI there is already automation present in the devfile-registry for using digests instead of tags. For example we have the `USE_DIGESTS` argument in the build scripts.

IMO the work here should look like this:
* [ ] Change the current references to digests manually
* [ ] Update the build scripts to use digests all the time (remove `USE_DIGESTS` and make it true)
* [ ] Add a `CONTRIBUTING.MD` to make it clear for new users how to contribute updates to the devfile-registry, and that they should use digests instead of tags
* [ ] Update the `README` to reflect the above

@nickboldt did I miss a case?

nickboldt:
>  Update the build scripts to use digests all the time (remove USE_DIGESTS and make it true)

Right now the scripts rely on seeing a tag and replacing that with the equivalent digest. 

If you hardcode the digests into the registry, how will the script know which NEWER digest to use?

Or are you expecting that ever new release of a sidecar container will be manually added to the plugin registry by someone who will find the latest digest and use that in their PR?

I suppose that works but... you could also just use container foo:2.4.6 and let the registry build convert that to foo@sha256:aasdasdasdasd for you.

ericwill:
Because we are to use tokens now, let's close this PR and open a new one when the new approach is decided.

PR #233, opened 2020-04-28
Enabling quarkus native compilation

sunix:
try it: http://che.openshift.io/factory?url=https://raw.githubusercontent.com/eclipse/che-devfile-registry/a1ea5f8915fad9b582be5c347b767ee555f14053/devfiles/quarkus/devfile.yaml

### DIFF
--- a/devfiles/quarkus/devfile.yaml
+++ b/devfiles/quarkus/devfile.yaml
@@ -59,7 +59,7 @@ commands:
         command: "./mvnw package"
         workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
   -
-    name: Start Dev mode (Hot reload + debug)
+    name: Start Development mode (Hot reload + debug)
     actions:
       -
         type: exec
--- a/devfiles/quarkus/devfile.yaml
+++ b/devfiles/quarkus/devfile.yaml
@@ -15,7 +15,7 @@ components:
     id: redhat/quarkus-java8/latest
   -
     type: dockerimage
-    alias: maven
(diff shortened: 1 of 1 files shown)


devfiles/quarkus/devfile.yaml
```diff
@@ -15,7 +15,7 @@ components:
     id: redhat/quarkus-java8/latest
   -
     type: dockerimage
-    alias: maven
+    alias: centos-quarkus-maven
```
ericwill, line 18: Why centos?

sunix: because it is the original name of the image being used. I would like to emphasis the fact that we are based on containers and although we are patching it we are using https://quay.io/repository/quarkus/centos-quarkus-maven

sunix: @ericwill ^^

devfiles/quarkus/devfile.yaml
```diff
@@ -25,35 +25,64 @@ components:
...
     actions:
       -
         type: exec
-        component: maven
+        component: centos-quarkus-maven
         command: "./mvnw package"
         workdir: ${CHE_PROJECTS_ROOT}/quarkus-quickstarts/getting-started
   -
-    name: Start Development mode
+    name: Start Dev mode (Hot reload + debug)
```
ericwill, line 62: Cleaner would be "Start Development Mode (Hot Reload + Debug)"

ericwill: Otherwise the PR looks good to me.

PR #238, opened 2020-05-13
Allow user to add TLS certificates to Java Keystore

tomgeorge:
### What does this PR do?

Fixes https://github.com/eclipse/che/issues/16920.

This adds a script that checks if `$JAVA_HOME` and the `cacerts` file exists, and if it does, sets the permissions of it to `664`, and then removes the script from the image.


tomgeorge:
Closing because this isn't necessary

PR #249, opened 2020-06-10
Quarkus devfile: add more memory for native compilation

sunix:
### What does this PR do?
Setting to 4927M: it looks like Quarkus 1.5.0-Final needs more memory to perform the native comp (tested on che.openshift.io). 
also skipping tests

### DIFF
--- a/devfiles/quarkus/devfile.yaml
+++ b/devfiles/quarkus/devfile.yaml
@@ -25,7 +25,7 @@ components:
           -Duser.home=/home/user"
       - name: MAVEN_OPTS
         value: $(JAVA_OPTS)  
-    memoryLimit: 4Gi
+    memoryLimit: 4927M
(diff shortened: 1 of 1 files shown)


apupier:
Iit seems that with a default account on che.openshift.io, the workspace is not starting because the total memory used is too high.

sunix:
I tried with a "normal account" (not redhat one) and it was working :(

JPinkney:
Originally I tested once and it worked fine. I've tested it 5 more times and it succeeded 4 times but failed once with out of memory

sunix:
@JPinkney was your out of memory when you perform the quarkus native compilation or workspace not able to start ?

JPinkney:
@sunix My workspace was able to start every time successfully the only issue was running the quarkus native compilation command

PR #250, opened 2020-06-11
Added docker buildx support in cico scripts

aditijadhav38:
### What does this PR do?
Added docker buildx support in cico scripts

### DIFF
--- a/cico_build_ci.sh
+++ b/cico_build_ci.sh
@@ -23,4 +23,6 @@ export SCRIPT_DIR
 load_jenkins_vars
 install_deps
 setup_environment
+check_buildx_support
 build_and_push
+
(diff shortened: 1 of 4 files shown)


AndrienkoAleksandr:
You need to fix shell check: https://github.com/eclipse/che-devfile-registry/pull/250/checks?check_run_id=762136014

AndrienkoAleksandr:
You almost fixed shell check. I think you can also use online validation using https://www.shellcheck.net It returns result very similar to our CI.

nickboldt:
As the cico scripts are no longer being maintained (centos CI is EOL), this PR will never be merged.

Instead, please submit a new PR with code that calls a PLATFORMS file and runs buildx like this:

* https://github.com/eclipse/che-plugin-registry/blob/master/.ci/sidecar-build-publish.sh#L31-L35

or like this:

* https://github.com/che-incubator/che-dashboard-next/blob/master/.github/workflows/dockerimage.yml#L35

PR #253, opened 2020-06-24
Add node-debug plugin into nodejs stacks with debug configuration

svor:

### What does this PR do?
Adds node-debug plugin into stacks with node debug configuration

Depends on https://github.com/eclipse/che-plugin-registry/pull/506

https://issues.redhat.com/browse/CRW-997

### DIFF
--- a/devfiles/nodejs-mongo/devfile.yaml
+++ b/devfiles/nodejs-mongo/devfile.yaml
@@ -13,6 +13,9 @@ components:
     type: chePlugin
     id: che-incubator/typescript/latest
     memoryLimit: 512Mi
+  - 
+    type: chePlugin
+    id: ms-vscode/node-debug2/latest
   -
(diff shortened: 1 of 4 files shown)


tsmaeder:
@svor if we _always_ need both extensions, would it make sense to include them in the same Che plugin?

svor:
@tsmaeder I don't see any problem to merge these plugins. But they can live separately and it is not necessarily to add node-debug plugin if you don't want to debug the application.
Should i create an issue for that, @ericwill WDYT?

tsmaeder:
@svor sorry, they _are_ in the same Che plugin in the newest version. I was trying to delete my comment, but Github's "eventual consistency" hid it, I guess ;-)

svor:
ah, i thought you meant node-debug + typescript-ls

PR #255, opened 2020-06-26
Fix offline build failures

monaka:
### What does this PR do?

Fix failures on `./build.sh --offline`.

This have two fixes.

1. Support `.source.commitId` on Devfiles.  It is used in `java-mysql-web-java-spring-petclinic`.
2. Specify `main` branch in `java-web-spring`. There is no `master` branch in the source repository.

None.

### DIFF
--- a/build/scripts/cache_projects.sh
+++ b/build/scripts/cache_projects.sh
@@ -34,7 +34,11 @@ function cache_project() {
   local sparse_checkout_dir="$4"
 
   rm -fr "$TEMP_REPO"
-  git clone "$repo" -b "$branch" --depth 1 "$TEMP_REPO" -q
+  git clone "$repo" -b "$branch" --depth 1 "$TEMP_REPO" -q || (
+     # In case branch is commitId.
+     git clone "$repo" "$TEMP_REPO" -q
+     echo "info: Fatal error by Git was recovered." > /dev/stderr
+  )
   pushd "$TEMP_REPO" &>/dev/null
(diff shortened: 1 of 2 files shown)


build/scripts/cache_projects.sh
```diff
@@ -34,7 +34,11 @@ function cache_project() {
   local sparse_checkout_dir="$4"
 
   rm -fr "$TEMP_REPO"
-  git clone "$repo" -b "$branch" --depth 1 "$TEMP_REPO" -q
+  git clone "$repo" -b "$branch" --depth 1 "$TEMP_REPO" -q || (
```
tsmaeder, line 37: If I understand this correctly, we're doing the second part if the first `git clone` fails for any reason. IMO, it would be cleaner to check explicitly whether `$branch` is empty. The first clone may fail for any number of reasons.

monaka: It may be misleading though in this case `$branch` is not empty. It contains the commit hash.
`git` can't clone with commit ID. This is the  reason I added the second part.

IMO, `$branch` is no always empty value. Since `$branch` is set as `master` before coming here if `$branch` was empty.

benoitf: I've opened https://github.com/eclipse/che-devfile-registry/pull/258 as alternate fix to really clone with the expected sha-1

benoitf:
taken over by https://github.com/eclipse/che-devfile-registry/pull/258

PR #262, opened 2020-07-02
fix(version): 8-->11

benoitf:
### What does this PR do?
Stacks was updated to use Java 11 and not java8, updating description

https://github.com/eclipse/che-devfile-registry/pull/257


monaka:
Nice catch. Thanks!

PR #264, opened 2020-07-07
fix(entrypoint): fix image error

benoitf:
### What does this PR do?

upstream tag image has been updated https://github.com/docker-library/httpd/commit/ea20f356a6f7d5b22793d0a51d82d6fccd08a6aa

it removes build dependencies but we were using bash transitively and it fails
add bash

good news: the image is A LOT smaller !

https://github.com/eclipse/che/issues/17345

Change-Id: I3f267521360ee0214bc0196561bd0c7ba164eacc


benoitf:
thanks @svor for the review

PR #268, opened 2020-07-22
Add sample project to Go stack

vitaliy-guliy:

### What does this PR do?
Adds another one sample project to Go stack.
Adds commands to build and run sample project.

If the `golang/go` plugin could not be found when creating a workspace from the devfile, the `id` should be replaced on `reference`.
```
-
  type: chePlugin
  # id: golang/go/latest
  reference: >-
    https://raw.githubusercontent.com/eclipse/che-plugin-registry/master/v3/plugins/golang/go/0.14.4/meta.yaml
  alias: go-plugin
  memoryLimit: 512Mi
  # memoryLimit: 1Gi
  env:
    - value: 'off'
      name: GO111MODULE
  preferences:
    go.lintTool: 'golangci-lint'
    go.lintFlags: '--fast'
    go.useLanguageServer: true
```

Solves issue https://github.com/eclipse/che/issues/13803

### DIFF
--- a/devfiles/go/devfile.yaml
+++ b/devfiles/go/devfile.yaml
@@ -9,10 +9,21 @@ projects:
     type: git
     location: https://github.com/golang/example.git
   clonePath: src/github.com/golang/example/
+-
+  name: golang-echo-realworld-example-app
+  source:
+    type: git
+    location: https://github.com/xesina/golang-echo-realworld-example-app.git
+  clonePath: src/github.com/xesina/golang-echo-realworld-example-app
+
 components:
 -
+  type: cheEditor
+  id: eclipse/che-theia/next
+  memoryLimit: 1Gi
+-
   type: chePlugin
-  id: ms-vscode/go/latest
+  id: golang/go/latest
   alias: go-plugin
   memoryLimit: 512Mi
   env:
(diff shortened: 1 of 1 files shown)


devfiles/go/devfile.yaml
```diff
@@ -25,51 +36,79 @@ components:
...
+-
+  name: 'Run current file'
   actions:
   - type: exec
     component: go-cli
     command: go get -d && go run ${file}
     workdir: ${fileDirname}
 -
-  name: Debug current file
+  name: 'Debug current file'
```
ericwill, line 111: Why the additional `'` marks?

devfiles/go/devfile.yaml
```diff
@@ -9,10 +9,21 @@ projects:
...
+-
+  type: cheEditor
+  id: eclipse/che-theia/next
+  memoryLimit: 1Gi
 -
   type: chePlugin
-  id: ms-vscode/go/latest
+  id: golang/go/latest
   alias: go-plugin
   memoryLimit: 512Mi
```
ericwill, line 28: Does the plugin need any more memory than before? Or can we leave it?

devfiles/go/devfile.yaml
vitaliy-guliy: It's a cosmetic decision. Just to be consistent with names of other command.

devfiles/go/devfile.yaml
vitaliy-guliy: The first I gave it 1Gb, but then reduced to the previous value.

vitaliy-guliy:
Yesterday I had some issues with creating a workspace from the devfile on https://che.openshift.io.
Today it works fine. It looks like the server was overloaded.

ericwill:
[ci-test-devfiles]

ericwill:
Looks like PR check is failing due to the renamed tasks.

ericwill:
@ScrewTSW where are these tests located so the task names can be updated?

PR #278, opened 2020-09-07
[Golang] Remove cheEditor from components.

monaka:
Signed-off-by: Masaki Muranaka <monaka@monami-ya.com>

### What does this PR do?

Removes `cheEditor` from components.
It is redundant because Che will choose its default editor from `CHE_FACTORY_DEFAULT__EDITOR` variable.

None.

### DIFF
--- a/devfiles/go/devfile.yaml
+++ b/devfiles/go/devfile.yaml
@@ -18,10 +18,6 @@ projects:
 
 components:
 -
-  type: cheEditor
-  id: eclipse/che-theia/next
-  memoryLimit: 1Gi
--
   type: chePlugin
   id: golang/go/latest
   alias: go-plugin


devfiles/go/devfile.yaml
```diff
@@ -17,10 +17,6 @@ projects:
   clonePath: src/github.com/xesina/golang-echo-realworld-example-app
 
 components:
--
-  type: cheEditor
-  id: eclipse/che-theia/next
-  memoryLimit: 1Gi
```
sleshchenko, line 23: I think it's here to override the default memory limit of the che-theia.

monaka: I agree it will override the limit. But I think there is no reason to do it.

@vitaliy-guliy It is added in #268 . Why is the memory limit expanded ?

vitaliy-guliy: I don't remember exactly, but it could because 512 Mb is not enough to run Che-Theia + plugins.

monaka: AFAIK, external plugins will run on the separated container.

sleshchenko: Well, whether plugins will run in the same containers or different depends on the plugins, but in this particular case, the only configured plugin is golang/go/latest and it's running in https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/golang/go/0.16.1/meta.yaml#L15...
Maybe theia itself needs more memory but if that sample works just fine after removing this override - then it's not clear why it's here.

monaka: Ah, my comment was not strict. I mean che-theia and language plugins won't share same container usually. At least in this case doesn't share.
So I suppose this override can be remove.

sleshchenko: @ericwill @vitaliy-guliy I think the final decision should be yours.

vitaliy-guliy: Do you know guys how to determine how much memory is allocated for Che-Theia in theia-ide container? (I don't know yet)
If it's nearly 512 Mb, I think it's better to give it a bit more memory to ensure Che-Theia will not fail when opening a couple of files.

sleshchenko: >  I think it's better to give it a bit more memory to ensure Che-Theia will not fail when opening a couple of files.

You think it makes sense only in go lang devfile or if such a problem exists - solve on another level and assign another default memory limit for theia? 
https://github.com/eclipse/che-plugin-registry/blob/master/v3/plugins/eclipse/che-theia/next/meta.yaml#L81

monaka: > another default memory limit for theia?

It's a considerable candidate.
And just my concern, the value of `requests.memory` is increased.
It may have an memory impact on multi-user environments since almost all users will use che-theia.

I'll agree to keep this override left if it is required. But currently there is no evidence.
How about trying remove it ?  And to revert (or patcing to che-theia) after the root cause is shown?

vitaliy-guliy: Well, let's remove it and then increase the memory for `eclipse/che-theia` if necessary.

PR #279, opened 2020-09-08
Update the command of installing dependencies in NodeJS+Yarn devfile

svor:

### What does this PR do?
Adds more memory for dev container;
Change the command to install dependencies from `yarn install` to `npm instaal`. Looks like yarn cannot install dependencies because the .node_modules was marked as symlink.

https://github.com/eclipse/che/issues/17453

### DIFF
--- a/arbitrary-users-patch/base_images
+++ b/arbitrary-users-patch/base_images
@@ -7,6 +7,7 @@ che-java11-gradle       gradle:6.2.1-jdk11
 che-java11-maven        maven:3.6.3-jdk-11
 che-java8-maven         maven:3.6.1-jdk-8
 che-nodejs10-community  node:10.16
+che-nodejs12-community  node:12.18
 che-nodejs10-ubi        registry.access.redhat.com/ubi8/nodejs-10
 che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
 che-php-7               quay.io/eclipse/che-php-base:7.4
--- a/devfiles/nodejs-yarn/devfile.yaml
+++ b/devfiles/nodejs-yarn/devfile.yaml
@@ -17,7 +17,7 @@ components:
     type: dockerimage
     alias: nodejs
     image: quay.io/eclipse/che-nodejs10-community:nightly
-    memoryLimit: 1Gi
+    memoryLimit: 1500Mi
     endpoints:
       - name: 'nodejs'
         port: 3000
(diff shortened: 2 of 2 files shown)


svor:
[ci-test-devfiles]

svor:
The command to install dependencies was changed from `yarn install` to `npm install` to avoid the next error:
screenshot-user-images githubusercontent com-2020 09 08-16_50_30
Since it is NodeJS+**Yarn** devfile is it OK to have `npm` in the command or should we do something similar as we have for che-theia and hosted Che https://github.com/eclipse/che-theia/blob/master/devfiles/hosted-che-dogfooding.devfile.yaml#L92 ?
@benoitf WDYT?

benoitf:
@svor I think you hit that issue only on che.openshift.io with glusterfs
could you try with ephemeral mode (attribute of devfile) ? (and then keep yarn install) as the issue is coming from che.openshift.io, not from the devfile

svor:
@benoitf Yes, you are right, I hit that issue only on che.openshift.io, and I tried all possible modes:
- In Ephemeral and Asynchronous i got **Disk quota exceeded**
- In Persistent: `error An unexpected error occurred: "EEXIST: file already exists, mkdir '/projects/react-web-app/node_modules/cosmiconfig'"`.
But when I tried to install dependencies by npm in Persistent mode, they were installed.

benoitf:
The issue is that the getting started is called `NodeJS Web Application based on Yarn` to differentiate with `on npm` so if we use npm instead of yarn it is not what is expected.

did you try by changing the path of node_modules folder with `--modules-folder /tmp/nodes_modules` when using persistent mode ?

or by using yarn v2 ?
`yarn set version berry `

svor:
The PR was updated to use Yarn 2
screenshot-che openshift io-2020 09 09-13_32_26

benoitf (approved):

benoitf:
@flacatus looks like digest check is failing

flacatus:
@svor @benoitf Is there a reason why I can not download `quay.io/eclipse/che-nodejs12-community:nightly`?

[flacatus@chehost-local chectl]$ docker pull quay.io/eclipse/che-nodejs12-community:nightly
Error response from daemon: unauthorized: access to the requested resource is not authorized

benoitf:
@flacatus it's brought by this PR

svor:
@flacatus it is not published yet, I've just added it in this PR https://github.com/eclipse/che-devfile-registry/pull/279/files#diff-d20815828764ec0fab8c6cafa5bdfd90R10

flacatus:
Oh, then I think it is ok to merge without diggest check if is not published.

JPinkney (approved):

PR #290, opened 2020-10-07
Switch nightly build to GH action

ericwill:
Run a GitHub action job everytime a change is merged, instead of using centos-ci.

See eclipse/che#18072

### DIFF
--- a/.github/workflows/build-jobs.yml
+++ b/.github/workflows/build-jobs.yml
@@ -16,9 +16,6 @@ on:
 
 jobs:
   nightly-build-push:
-    env:
-      QUAY_ECLIPSE_CHE_USERNAME: ${{ secrets.QUAY_USERNAME }}
-      QUAY_ECLIPSE_CHE_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
     runs-on: ubuntu-latest
     steps:
       - name: Clone source code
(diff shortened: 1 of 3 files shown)


ericwill:
Still WIP as the secrets need to be added to the repo.

ericwill:
> Still WIP as the secrets need to be added to the repo.

Added the secrets, the PR is ready for review.

.github/workflows/pr-jobs.yml
```diff
@@ -1,3 +1,12 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
 name: CI
```
JPinkney, line 10: Should we change the name of this now that this file was renamed from main -> pr-jobs?

ericwill: Good point, I made it more clear

JPinkney (approved):

ericwill:
I just realized that the script (still in this PR) installs some stuff specific to centos, which I will remove.

benoitf (commented):
forgot to hit send yesterday

.github/workflows/build-jobs.yml
```diff
@@ -0,0 +1,30 @@
...
+      QUAY_ECLIPSE_CHE_PASSWORD: ${{ secrets.QUAY_DOCKER_PASSWORD }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone source code
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+
+      - name: Build and push image
+        run: /bin/bash cico_build_nightly.sh
```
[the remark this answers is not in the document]
ericwill: We shouldn't get rid of them until all build jobs have been moved to GH actions, but I abstracted away what is needed for this PR

benoitf (approved):
Let's try

ericwill:
It worked -- the job took 18m 45s, almost 15 of which were due to the base image builds. Maybe these can be built only on merge on path `arbitrary-users-patch/*` and then build them nightly in a cron-job.

PR #294, opened 2020-10-13
Update devfiles to user 3.8.6 version of python stuff

tsmaeder:
### What does this PR do?
Update python 3.8.6 and a more recent version of vscode-python

https://github.com/eclipse/che/issues/16183

related PRs: 
* https://github.com/eclipse/che-devfile-registry/pull/294
* https://github.com/eclipse/che-plugin-registry/pull/641
* https://github.com/che-samples/django-realworld-example-app/pull/1

### DIFF
--- a/devfiles/python-django/devfile.yaml
+++ b/devfiles/python-django/devfile.yaml
@@ -7,7 +7,6 @@ projects:
     source:
       type: git
       location: 'https://github.com/che-samples/django-realworld-example-app'
-      commitId: 0d425480c1b28d6063c255e7c4fa869eb1709d85
 components:
   -
     type: chePlugin
--- a/devfiles/python-django/devfile.yaml
+++ b/devfiles/python-django/devfile.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: 1.0.0
 metadata:
   generateName: python-django-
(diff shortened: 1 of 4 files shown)


devfiles/python-django/devfile.yaml
```diff
@@ -77,3 +77,11 @@ commands:
             ]
           }]
         }
+    - 
+      name: set up venv
```
ericwill, line 81: When I tested this, the devfile editor in the Getting Started area would complain about this being after the `debug` task. When I moved it before the `debug` task, all was well. :man_shrugging:

devfiles/python-django/devfile.yaml
```diff
@@ -7,7 +6,8 @@ projects:
     name: django-realworld-example-app
     source:
       type: git
-      location: "https://github.com/che-samples/django-realworld-example-app"
+      location: 'https://github.com/che-samples/django-realworld-example-app'
+      commitId: 0d425480c1b28d6063c255e7c4fa869eb1709d85
```
ericwill, line 10: This might need to change once the commit has been merged, I'm not sure if GH respects commit ID's the way it ought to.

ericwill (commented):
It might also be nice to order the tasks somehow -- I knew the correct order because I'm familiar with Django/Python, but a new user might not.

devfiles/python-django/devfile.yaml
tsmaeder: Actually, the indentation is wrong. YAML is crap!

tsmaeder:
> It might also be nice to order the tasks somehow

I agree, but it would make sense to make a pass over everything and fix it. Also: if we ever have more than 9 steps, ordering is going to be broken.

ericwill (approved):

PR #321, opened 2021-01-05
Multiarch build support by docker buildx

Prabhav-Thali:
### What does this PR do?
Enables docker buildx support in GitHub Actions job to build and publish multiarch docker images. All the workflows (nightly build, PR check, and release) are updated to have multiarch support. 

This refers to [#17124](https://github.com/eclipse/che/issues/17124). As per @nickboldt's [comment](https://github.com/eclipse/che-devfile-registry/pull/250#issuecomment-740180491) on #250, raised this PR to add multiarch support.

### DIFF
--- a/.github/workflows/nightly-build-publish.yml
+++ b/.github/workflows/nightly-build-publish.yml
@@ -19,9 +19,22 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Clone source code
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
           fetch-depth: 1
+
+      - name: Prepare
+        id: prep
+        run: |
+          SHORT_SHA1=$(git rev-parse --short HEAD)
+          echo ::set-output name=short_sha1::${SHORT_SHA1}
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+        
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
 
       - name: Login to quay.io
         uses: docker/login-action@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -46,13 +61,15 @@ jobs:
...
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./build/dockerfiles/Dockerfile
+          target: registry
+          build-args: |
+            PATCHED_IMAGES_TAG=${{ steps.prep.outputs.version }}
+          platforms: linux/amd64,linux/s390x
+          tags: quay.io/eclipse/${IMAGE}:${{ steps.prep.outputs.version }},quay.io/eclipse/${IMAGE}:${{ steps.prep.outputs.short_sha1 }}
```
ericwill, line 73: `${IMAGE}` should be `che-devfile-registry`. Currently it will evaluate to nothing.

Prabhav-Thali: That was a miss from my side. Implemented the change, please review.

ericwill (approved):

ericwill:
Let's give it a shot

.github/workflows/nightly-build-publish.yml
```diff
@@ -37,10 +50,12 @@ jobs:
...
-          docker push quay.io/eclipse/che-devfile-registry:nightly
-          docker tag che-devfile-registry quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
-          docker push quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./build/dockerfiles/Dockerfile
+          target: registry
+          platforms: linux/amd64,linux/s390x
```
benoitf, line 59: we're missing 2 arches there: ppc64le and arm64

.github/workflows/pr-checks.yml
```diff
@@ -39,38 +41,45 @@ jobs:
...
-    - name: Cache docker layers
-      run: |
-        mkdir -p ./caches
-        docker save -o ./caches/app.tar app
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./build/dockerfiles/Dockerfile
+        target: registry
+        platforms: linux/amd64,linux/s390x
```
benoitf, line 70: looks like platforms should be a constant on top of the file to avoid copy/paste

.github/workflows/pr-checks.yml
```diff
@@ -39,38 +41,45 @@ jobs:
...
-        
-    - name: Load docker layer cache
-      run: |
-        set +o pipefail
-        docker load -i ./caches/app.tar | true
+        path: /tmp/.buildx-cache
+        key: v2-${{ github.head_ref }}
     
+    - name: 'Docker Prepare'
+      run: docker image prune -a -f
```
benoitf, line 62: why do we need to remove other images ? we're missing some spaces ?
if not we could keep them to not slow down

.github/workflows/pr-checks.yml
Prabhav-Thali: @benoitf - Thanks, Raised #325 to address this review comment.

.github/workflows/pr-checks.yml
Prabhav-Thali: @benoitf Had referred to che-dashboard .yml [file](https://github.com/eclipse/che-dashboard/blob/381ff548a9fff3537f1a29ce8e9b228f6c145338/.github/workflows/pr.yaml#L49) for buildx implementation in GA. However, I have removed prune step in #325

PR #325, opened 2021-01-11
Adds platforms file for buildx

Prabhav-Thali:
### What does this PR do?
Adds PLATFORMS file rather than mentioning platforms in multiple .yml files. Also, adds 2 more platforms, i,e ppc64le and arm64.

This is a follow-up PR to implement review comments provided by @benoitf on #321 .

### DIFF
--- a/PLATFORMS
+++ b/PLATFORMS
@@ -1,0 +1,1 @@
+linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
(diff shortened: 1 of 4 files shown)


benoitf (approved):
thanks @Prabhav-Thali  👍

PLATFORMS
```diff
@@ -0,0 +1 @@
+linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
```
benoitf, line 1: very minor: newline should be added

Prabhav-Thali: Implemented

ericwill (approved):

PR #333, opened 2021-02-09
Use generateName for bash devfile to match others

amisevsk:
### What does this PR do?
Change the bash sample to use `generateName`, matching other sample devfiles in this repo.


amisevsk:
This fixes an issue where the bash sample can't be used twice by the same user without deleting the previous workspace.

sunix (approved):

JPinkney (approved):

benoitf (approved):

PR #12, opened 2019-06-13
More devfiles

sleshchenko:
### What does this PR do?
This PR adds projects along with commands to existing devfiles:

| done | devfile | sample |
| ----- | ----- | ------ |
| :heavy_check_mark:  | `java-maven` | https://github.com/che-samples/console-java-simple |
| :heavy_check_mark: | `java-gradle` | https://github.com/che-samples/console-java-simple |
| :heavy_check_mark: | `apache-camel-springboot` | https://github.com/jboss-fuse/fuse-rest-http-booster |
| :heavy_check_mark: | `dotnet` (c#) | https://github.com/che-samples/dotnet-web-simple.git |
| :heavy_check_mark: | `python` | https://github.com/che-samples/console-python3-simple.git |
| :heavy_check_mark: | `python-django` | https://github.com/gothinkster/django-realworld-example-app |

Notes:
All devfiles and commands are tested but the following issues are faced:
1. downloading maven dependencies for `apache-camel-springboot` takes an age. Need to check and maybe use a lighter sample project.
Screenshot_20190614_140101
2. python-django is working but error message is shown
Screenshot_20190614_114424
https://github.com/eclipse/che/issues/13562

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/13529

PR #336, opened 2021-02-10
Added multiarch support by docker buildx

bivasda1:
Signed-off-by: Bivas Das <bivasda1@in.ibm.com>

### What does this PR do?
Enables docker buildx support in GitHub Actions job to build and publish multiarch docker images. All the workflows (nightly build, PR check, and release) are updated to have multiarch support.

This refers to https://github.com/eclipse/che-devfile-registry/pull/321

### Reviewers
@ericwill Could you please review this PR

### DIFF
--- a/.github/workflows/nightly-build-publish.yml
+++ b/.github/workflows/nightly-build-publish.yml
@@ -18,29 +18,55 @@ jobs:
   nightly-build-publish:
     runs-on: ubuntu-latest
     steps:
-      - name: Clone source code
+      - 
+        name: Clone source code
         uses: actions/checkout@v1
         with:
           fetch-depth: 1
-
-      - name: Login to quay.io
+      - 
+        name: Prepare
+        id: prep
+        run: |
+          SHORT_SHA1=$(git rev-parse --short HEAD)
+          echo ::set-output name=short_sha1::${SHORT_SHA1}
+          VERSION=$(head -n 1 VERSION)
+          echo ::set-output name=version::${VERSION}
+          IMAGE=che-devfile-registry
+          echo ::set-output name=image::${IMAGE}
+          PLATFORMS=$(cat PLATFORMS)
+          echo ::set-output name=platforms::${PLATFORMS}
+      -
+        name: "Set up QEMU"
+        uses: docker/setup-qemu-action@v1
+      -
+        name: "Set up Docker Buildx"
+        uses: docker/setup-buildx-action@v1
+      -
+        name: "Docker quay.io Login"
         uses: docker/login-action@v1
         with:
           registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
+      - 
+        name: Login to docker.io
+        uses: docker/login-action@v1
+        with:
+          registry: docker.io
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
       - name: Build and push base images
         run: /bin/bash arbitrary-users-patch/build_images.sh --push
 
       - name: Build and push happy path image
         run: /bin/bash arbitrary-users-patch/happy-path/build_happy_path_image.sh --push
-
-      - name: Build and push devfile image
-        run: |
-          SHORT_SHA1=$(git rev-parse --short HEAD)
-          docker build -t che-devfile-registry -f ./build/dockerfiles/Dockerfile --target registry .
-          docker tag che-devfile-registry quay.io/eclipse/che-devfile-registry:nightly
-          docker push quay.io/eclipse/che-devfile-registry:nightly
-          docker tag che-devfile-registry quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
-          docker push quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
+      -
+        name: "Build and push"
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./build/dockerfiles/Dockerfile
+          platforms: ${{ steps.package.outputs.content }}
+          tags:  quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.version }},quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.short_sha1 }}
+          push: true 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,29 +17,51 @@ on:
         required: true
 
 jobs:
-  release-build-publish:
+  build:
+    name: Create Che Devfile Registry Release
     runs-on: ubuntu-latest
     steps:
-      - name: Clone source code
+      -
+        name: "Checkout  source code"
         uses: actions/checkout@v1
         with:
           fetch-depth: 1
           ref: release
-
-      - name: Login to Quay.io
-        uses: azure/docker-login@v1
+      - 
+        name: Prepare
+        id: prep
+        run: |
+          SHORT_SHA1=$(git rev-parse --short HEAD)
+          echo ::set-output name=short_sha1::${SHORT_SHA1}
+          VERSION=$(head -n 1 VERSION)
+          echo ::set-output name=version::${VERSION}
+          IMAGE=che-devfile-registry
+          echo ::set-output name=image::${IMAGE}
+          PLATFORMS=$(cat PLATFORMS)
+          echo ::set-output name=platforms::${PLATFORMS}
+      -
+        name: "Set up QEMU"
+        uses: docker/setup-qemu-action@v1
+      -
+        name: "Set up Docker Buildx"
+        uses: docker/setup-buildx-action@v1
+      -
+        name: "Docker quay.io Login"
+        uses: docker/login-action@v1
         with:
-          login-server: quay.io
+          registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
-
-      - name: Login to Docker Hub
-        uses: azure/docker-login@v1
+      - 
+        name: Login to docker.io
+        uses: docker/login-action@v1
         with:
+          registry: docker.io
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
-
-      - name: Build and push patched base images and happy path image
+          
+      - 
+        name: Build and push patched base images and happy path image
         run: |
           export GITHUB_TOKEN=${{ secrets.CHE_BOT_GITHUB_TOKEN }}
           ./make-release.sh --version ${{ github.event.inputs.version}} --trigger-release
--- a/PLATFORMS
+++ b/PLATFORMS
@@ -1,0 +1,1 @@
+linux/amd64,linux/ppc64le,linux/arm64
--- a/arbitrary-users-patch/build_images.sh
+++ b/arbitrary-users-patch/build_images.sh
@@ -20,6 +20,8 @@ REGISTRY=${REGISTRY:-${DEFAULT_REGISTRY}}
 ORGANIZATION=${ORGANIZATION:-${DEFAULT_ORGANIZATION}}
 TAG=${TAG:-${DEFAULT_TAG}}
 
+PLATFORMS=$(cat arbitrary-users-patch/PLATFORMS)
+
 NAME_FORMAT="${REGISTRY}/${ORGANIZATION}"
 
 PUSH_IMAGES=false
@@ -32,7 +34,7 @@ while read -r line; do
   base_image_name=$(echo "$line" | tr -s ' ' | cut -f 1 -d ' ')
   base_image=$(echo "$line" | tr -s ' ' | cut -f 2 -d ' ')
   echo "Building ${NAME_FORMAT}/${base_image_name}:${TAG} based on $base_image ..."
-  docker build -t "${NAME_FORMAT}/${base_image_name}:${TAG}" --no-cache --build-arg FROM_IMAGE="$base_image" "${SCRIPT_DIR}"/ | cat
+  docker buildx build --platform "$PLATFORMS" -t "${NAME_FORMAT}/${base_image_name}:${TAG}" --no-cache --build-arg FROM_IMAGE="$base_image" "${SCRIPT_DIR}"/ | cat
   if ${PUSH_IMAGES}; then
     echo "Pushing ${NAME_FORMAT}/${base_image_name}:${TAG}" to remote registry
     docker push "${NAME_FORMAT}/${base_image_name}:${TAG}" | cat
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -41,6 +41,20 @@ jobs:
     steps:
     - uses: actions/checkout@v1
 
+    - 
+      name: Prepare
+      id: prep
+      run: |
+        PLATFORMS=$(cat PLATFORMS)
+        echo ::set-output name=platforms::${PLATFORMS}
+        
+    - name: Set up Docker Buildx
+      id: buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+      
     - name: Cache docker layers
       uses: actions/cache@v1
       with:
@@ -53,23 +67,24 @@ jobs:
         docker load -i ./caches/app.tar | true
     
     - name: Build devfile registry
-      run: |
-        docker build \
-          --cache-from=app \
-          -t app \
-          -f ./build/dockerfiles/Dockerfile \
-          --target registry .
-           
-    - name: Cache docker layers
-      run: |
-        mkdir -p ./caches
-        docker save -o ./caches/app.tar app
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./build/dockerfiles/Dockerfile
+        target: registry
+        platforms: ${{ steps.prep.outputs.platforms }}
+        tags: app
+        cache-from: "type=local,src=/tmp/.buildx-cache"
+        cache-to: "type=local,dest=/tmp/.buildx-cache"
+        push: false
     
     - name: Build offline devfile registry
-      run: |
-        docker build \
-          --cache-from=app \
-          -t app \
-          -f ./build/dockerfiles/Dockerfile \
-          --target offline-registry .
-   
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: ./build/dockerfiles/Dockerfile
+        target: offline-registry
+        platforms: ${{ steps.prep.outputs.platforms }}
+        tags: app
+        cache-from: "type=local,src=/tmp/.buildx-cache"
+        push: false
--- a/arbitrary-users-patch/PLATFORMS
+++ b/arbitrary-users-patch/PLATFORMS
@@ -1,0 +1,1 @@
+linux/amd64,linux/ppc64le,linux/arm64
--- a/arbitrary-users-patch/happy-path/PLATFORMS
+++ b/arbitrary-users-patch/happy-path/PLATFORMS
@@ -1,0 +1,1 @@
+linux/amd64,linux/ppc64le,linux/arm64
--- a/arbitrary-users-patch/happy-path/build_happy_path_image.sh
+++ b/arbitrary-users-patch/happy-path/build_happy_path_image.sh
@@ -20,6 +20,8 @@ REGISTRY=${REGISTRY:-${DEFAULT_REGISTRY}}
 ORGANIZATION=${ORGANIZATION:-${DEFAULT_ORGANIZATION}}
 TAG=${TAG:-${DEFAULT_TAG}}
 
+PLATFORMS=$(cat arbitrary-users-patch/happy-path/PLATFORMS)
+
 NAME_FORMAT="${REGISTRY}/${ORGANIZATION}"
 
 PUSH_IMAGES=false
@@ -28,7 +30,7 @@ if [ "$1" == "--push" ]; then
 fi
 
 # Build image for happy-path tests with precashed mvn dependencies
-docker build -t "${NAME_FORMAT}/happy-path:${TAG}" --no-cache --build-arg TAG="${TAG}" "${SCRIPT_DIR}"/  | cat
+docker buildx build --platform "$PLATFORMS" -t "${NAME_FORMAT}/happy-path:${TAG}" --no-cache --build-arg TAG="${TAG}" "${SCRIPT_DIR}"/  | cat
 if ${PUSH_IMAGES}; then
     echo "Pushing ${NAME_FORMAT}/happy-path:${TAG}" to remote registry
     docker push "${NAME_FORMAT}/happy-path:${TAG}" | cat
--- a/make-release.sh
+++ b/make-release.sh
@@ -38,7 +38,7 @@ performRelease()
   VERSION=$(head -n 1 VERSION)
   IMAGE=che-devfile-registry
   DOCKERFILE_PATH=./build/dockerfiles/Dockerfile
-  docker build -t ${IMAGE} -f ${DOCKERFILE_PATH} --build-arg PATCHED_IMAGES_TAG="${VERSION}" --target registry .
+  docker buildx build --platform "$PLATFORMS" -t ${IMAGE} -f ${DOCKERFILE_PATH} --build-arg PATCHED_IMAGES_TAG="${VERSION}" --target registry .
   docker tag ${IMAGE} "quay.io/eclipse/${IMAGE}:${SHORT_SHA1}"
   docker push "quay.io/eclipse/${IMAGE}:${SHORT_SHA1}"
   docker tag ${IMAGE} "quay.io/eclipse/${IMAGE}:${VERSION}"


bivasda1:
FYI - @nickboldt - Follow up PR as per comments here - https://issues.redhat.com/browse/CRW-1475?focusedCommentId=15721358&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15721358

.github/workflows/release.yml
```diff
@@ -1,9 +1,3 @@
-#
-# Copyright (c) 2020 Red Hat, Inc.
```
ericwill, line 2: Copyright header needs to stay, but the year should be updated

.github/workflows/release.yml
```diff
@@ -13,51 +7,60 @@ on:
...
+      -
+        name: "Docker quay.io Login"
+        uses: docker/login-action@v1
         with:
-          login-server: quay.io
+          registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
-
-      - name: Login to Docker Hub
```
ericwill, line 41: We need to log into both quay.io and dockerhub. The reason is that we push to quay, and we need to authenticate with dockerhub to prevent rate limiting.

.github/workflows/release.yml
```diff
@@ -13,51 +7,60 @@ on:
...
-  release-build-publish:
+  build:
+    name: Create Che Devfile Registry Release
     runs-on: ubuntu-latest
     steps:
-      - name: Clone source code
-        uses: actions/checkout@v1
+      -
+        name: "Checkout  source code"
+        uses: actions/checkout@v2
```
ericwill, line 26: Any particular reason to use `v2` over `v1`?

.github/workflows/release.yml
```diff
@@ -13,51 +7,60 @@ on:
...
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_PASSWORD }}
-
-      - name: Build and push patched base images and happy path image
+      - 
+        name: Build and push patched base images and happy path image
         run: |
-          export TAG=$(head -n 1 VERSION)
           /bin/bash arbitrary-users-patch/happy-path/build_happy_path_image.sh --push
```
ericwill, line 56: Both the happy path image and the base images need to be built with multi-arch support as well, otherwise the devfiles inside the registry will be no-op. I would suggest modifying the build scripts and adding `PLATFORMS` files.

.github/workflows/nightly-build-publish.yml
```diff
@@ -36,11 +53,17 @@ jobs:
...
+      - 
+        name: Build and push patched base images and happy path image
         run: |
-          SHORT_SHA1=$(git rev-parse --short HEAD)
-          docker build -t che-devfile-registry -f ./build/dockerfiles/Dockerfile --target registry .
-          docker tag che-devfile-registry quay.io/eclipse/che-devfile-registry:nightly
-          docker push quay.io/eclipse/che-devfile-registry:nightly
-          docker tag che-devfile-registry quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
-          docker push quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
+          /bin/bash arbitrary-users-patch/happy-path/build_happy_path_image.sh --push
```
ericwill, line 59: Both the happy path image and the base images need to be built with multi-arch support as well, otherwise the devfiles inside the registry will be no-op. I would suggest modifying the build scripts and adding `PLATFORMS` files.

ericwill (changes requested):
A few changes are required, please see comments inline. Also, please test to make sure that the [regression](https://github.com/eclipse/che-devfile-registry/actions/runs/515108378) from the last attempt doesn't happen this time.

.github/workflows/release.yml
bivasda1: @ericwill no as such, Please let me know do you require v1 over v2?

ericwill: I don't think so, I see we are using `fetch-depth` so it shouldn't make a difference.

bivasda1:
@ericwill Can you please review this PR, I have changed Accordingly
 
**Below change requests are updated**
- Copyright header and  also year is Updated
- Enabled log into both quay.io and dockerhub
- Enabled multiarch build using buildx for Both the happy path image and the base images 
- Added Platformfile

PLATFORMS
```diff
@@ -0,0 +1 @@
+linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
```
bivasda1, line 1: Platform file Added

arbitrary-users-patch/build_images.sh
```diff
@@ -32,7 +34,7 @@ while read -r line; do
   base_image_name=$(echo "$line" | tr -s ' ' | cut -f 1 -d ' ')
   base_image=$(echo "$line" | tr -s ' ' | cut -f 2 -d ' ')
   echo "Building ${NAME_FORMAT}/${base_image_name}:${TAG} based on $base_image ..."
-  docker build -t "${NAME_FORMAT}/${base_image_name}:${TAG}" --no-cache --build-arg FROM_IMAGE="$base_image" "${SCRIPT_DIR}"/ | cat
+  docker buildx build --platform "$PLATFORMS" -t "${NAME_FORMAT}/${base_image_name}:${TAG}" --no-cache --build-arg FROM_IMAGE="$base_image" "${SCRIPT_DIR}"/ | cat
```
bivasda1, line 37: Enabled docker buildx for multiarch build

.github/workflows/release.yml
```diff
@@ -1,3 +1,9 @@
+#
```
bivasda1, line 1: copyright and year is updated

bivasda1 (commented):
Change requests are updated

PLATFORMS
ericwill: I would prefer this file be added in `arbitrary-users-patch` and `arbitrary-users-patch/happy-path/`, and adjust the scripts accordingly. That way the build script can be run from the repository root, or from within the directories themselves.

bivasda1:
@ericwill added Platform file in **arbitrary-users-patch and arbitrary-users-patch/happy-path** location, accordingly modified the script.

ericwill (dismissed):

ericwill:
@bivasda1 please rebase the PR and let's merge it next week. There is a `7.26.2` release next week so that will be a good test run.

nickboldt:
@mkuznyetsov FYI ^

@bivasda1 if rebased and ready by EOD Tuesday, we can include in Wednesday's release. If not... will slip to the week after. 

Note too that we've been having issues with qemu emulated s390x buildx builds, so you might want to only include ppc64le in  this PR as the s390x build might make it harder to validate.

bivasda1:
@nickboldt I will remove **s390x** form Platform files and will rebase as quickly as possible.

bivasda1:
@nickboldt @ericwill I have rebased.

.github/workflows/release.yml
```diff
@@ -13,51 +13,69 @@ on:
...
-          docker tag ${IMAGE} quay.io/eclipse/${IMAGE}:${VERSION}
-          docker push quay.io/eclipse/${IMAGE}:${VERSION}
+      -
+        name: "Build and push"
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./build/dockerfiles/Dockerfile
+          platforms: ${{  steps.prep.outputs.platforms }}
+          tags:  quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.version }},quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.short_sha1 }},quay.io/eclipse/${{ steps.prep.outputs.image }}:${{  github.event.inputs.version }}
```
nickboldt, line 80: you're pushing the same image w/ 3 tags here:
*           quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.version }},
*           quay.io/eclipse/${{ steps.prep.outputs.image }}:${{  github.event.inputs.version }}
*           quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.short_sha1 }},

should it be only two pushes for prep.outputs.version and short_sha1? Not sure why we would want the submitted version 7.y.z AND version from VERSION file AND the short sha from the tip of the checked out branch. 

@mkuznyetsov WDYT?

bivasda1: @nickboldt 

> should it be only two pushes for prep.outputs.version and short_sha1?
Yes
```
prep.outputs.version-->7.27.0-SNAPSHOT
event.inputs.version-->7.27.0 
```
I was trying to release proper version, so that `event.inputs.version` I have added. Please let me know should I remove `event.inputs.version`?

I will remove spaces from `${{ github.event.inputs.version }}`

.github/workflows/release.yml
```diff
@@ -13,51 +13,69 @@ on:
...
-          docker tag ${IMAGE} quay.io/eclipse/${IMAGE}:${VERSION}
-          docker push quay.io/eclipse/${IMAGE}:${VERSION}
+      -
+        name: "Build and push"
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./build/dockerfiles/Dockerfile
+          platforms: ${{  steps.prep.outputs.platforms }}
+          tags: quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.version }},quay.io/eclipse/${{ steps.prep.outputs.image }}:${{ steps.prep.outputs.short_sha1 }}
```
bivasda1, line 80: @nickboldt I have removed ` quay.io/eclipse/${{ steps.prep.outputs.image }}:${{  github.event.inputs.version }}` and rebased

.github/workflows/release.yml
nickboldt: we don't release -SNAPSHOT versions to quay.

https://quay.io/repository/eclipse/che-devfile-registry?tab=tags

Just nightly + sha tag and release (7.yy.z) + sha tag.

.github/workflows/release.yml
nickboldt: > outputs.version-->7.27.0-SNAPSHOT

That's the wrong version to release. We would want 7.27.0, not the -SNAPSHOT version. Releases are for releases, nightly is for snapshots. :D

.github/workflows/release.yml
```diff
@@ -13,51 +13,69 @@ on:
...
-
-      - name: Login to Quay.io
-        uses: azure/docker-login@v1
+      - 
+        name: Prepare
+        id: prep
+        run: |
+          SHORT_SHA1=$(git rev-parse --short HEAD)
+          echo ::set-output name=short_sha1::${SHORT_SHA1}
+          VERSION=$(head -n 1 VERSION)
```
nickboldt, line 41: If this value != `inputs.version` we should use inputs.version.

In fact I'm not sure why we need BOTH a computed value (7.27.0-SNAPSHOT) and an input value (7.27.0). 

I would expect that if the input is missing, we compute it... but I believe part of the release process is to SET the value in the VERSION file, so there's no point reading the file before we set the value. :)  

Instead I'd just take the value from the input as it should be the same as the VERSION file, once the release is complete. 

Correct @mkuznyetsov ?

bivasda1: @nickboldt do you expect ci-multiarch workflows should be formatted https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml#L91-L115 referenced to https://github.com/eclipse/che-dashboard/pull/119#issuecomment-783374087

nickboldt:
> do you expect ci-multiarch workflows should be formatted

Since I don't own this project, I'll let the TL comment... but my $0.02 is that it seems like from a project management perspective it might be better/more efficient to have:

* 1 GH action that can handle all three (four) arches, rather than 
* 1 GH action action for x64 and ppc64 (and aarch64), while a travis script separately handles s390x.

.github/workflows/release.yml
```diff
@@ -13,51 +13,69 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'release version in format 7.y.z'
+        description: 'The version that is going to be released. Should be in format 7.y.z'
         required: true
+        default: '7.y.z'
```
mkuznyetsov, line 18: I'd suggest having an empty default value, so we wouldn't have accidental workflow runs with `7.y.z` taken as a default parameter.

.github/workflows/release.yml
bivasda1: > Instead I'd just take the value from the input as it should be the same as the VERSION file, once the release is complete.

@nickboldt can we modify the version file `7.27.0-SNAPSHOT->7.27.0` apply the condition if `[vesion==input.version]` then release the specific version

mkuznyetsov: @nickboldt as of now, make-release.sh is called first from another workflow, which makes changes to VERSION file, and then this workflow does the rest. Here it reads the VERSION file, in case it was triggered by the push to the release branch, which has no inputs

Because it was slightly confusing, I made a PR (about which I've forgot 😃 )
https://github.com/eclipse/che-devfile-registry/pull/338/files
it makes `make-release.sh` will be called from this workflow

bivasda1:
> > do you expect ci-multiarch workflows should be formatted https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml#L91-L115 referenced to eclipse/che-dashboard#119
> 
> Since I don't own this project, I'll let the TL comment... but my $0.02 is that it seems like from a project management perspective it might be better/more efficient to have:
> 
> * 1 GH action that can handle all three (four) arches, rather than
> * 1 GH action action for x64 and ppc64 (and aarch64), while a travis script separately handles s390x.



> > do you expect ci-multiarch workflows should be formatted
> 
> Since I don't own this project, I'll let the TL comment... but my $0.02 is that it seems like from a project management perspective it might be better/more efficient to have:
> 
> * 1 GH action that can handle all three (four) arches, rather than
> * 1 GH action action for x64 and ppc64 (and aarch64), while a travis script separately handles s390x.

@ericwill Could you please share your thoughts @nickboldt suggestions

bivasda1:
@nickboldt Could you please review this PR,I think Visioning issue is resolved  now.

ericwill (commented):
Please remove this file: `.github/workflows/.nightly-build-publish.yml.swp`

ericwill:
> > > do you expect ci-multiarch workflows should be formatted https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml#L91-L115 referenced to [eclipse/che-dashboard#119](https://github.com/eclipse/che-dashboard/pull/119)
> > 
> > 
> > Since I don't own this project, I'll let the TL comment... but my $0.02 is that it seems like from a project management perspective it might be better/more efficient to have:
> > 
> > * 1 GH action that can handle all three (four) arches
> >   rather than
> > * 1 GH action action for x64 and ppc64 (and aarch64), while a travis script separately handles s390x.
> 
> @ericwill Could you please share your thoughts @nickboldt suggestions

I agree with the first option: `1 GH action that can handle all three (four) arches`

bivasda1:
> Please remove this file: `.github/workflows/.nightly-build-publish.yml.swp`

Opps,I will remove that.Thanks

bivasda1:
> > Please remove this file: `.github/workflows/.nightly-build-publish.yml.swp`
> 
Could u please review the PR I have  deleted `.github/workflows/.nightly-build-publish.yml.swp
`

.github/workflows/release.yml
ericwill: So which approach are we taking? Right now as it stands, the `7.27.0-SNAPSHOT` problem won't happen because the workflow is triggered after the file is changed. But we can have this PR wait until https://github.com/eclipse/che-devfile-registry/pull/338/files is merged. What do you guys think? cc @nickboldt @mkuznyetsov

nickboldt: PR 338 is ready to merge, after fixing all the shellcheck complaints. https://github.com/eclipse/che-devfile-registry/pull/338/files

ericwill: I've merged it, @bivasda1 please adjust the PR to make sure `-SNAPSHOT` isn't pushed as part of the tag

nickboldt:
please rebase, you're out of date again :(

bivasda1:
@nickboldt Rebased and Added buildx support `make-release.sh`

ericwill (approved):
Let's give it a shot

nickboldt:
seems legit for x and power arches but this won't work for s390x and will need to be refactored in future.

+1 for pushing this for now, with the caveat that we'll need something like https://github.com/eclipse/che-dashboard/pull/177 to support pushing each per-arch to quay independently, then combining them into a single multi-arch digest/manifest.

nickboldt (approved):
+1, but won't work for s390x.

ArvinB:
Just commenting...here is a bash script very similar to what I do for building multi-arch images. As @nickboldt mentioned from https://github.com/eclipse/che-dashboard/pull/177 that there is a risk of building for `s390x`. My issue primarily is that the CentOS repo's do not contain the packages for `s390x`. I don't have enough knowledge yet on which repo(s) to include, so for now I disable them all so that basically the yum and dnf updates all just pass on through.

```
registry=quay.io
repo=some/path
version=1.0.0
platforms=("amd64" "s390x")

tags=""
for platform in ${platforms[*]}
do
  echo "Building for linux/$platform"
  tag=$registry/$repo:${version}_$platform
  docker buildx build --load --platform linux/$platform -t $tag .
  docker push $tag
  tags="${tags} ${tag}"
done

docker manifest create $registry/$repo:$version $tags
docker manifest push $registry/$repo:$version
docker buildx imagetools inspect $registry/$repo:$version
```

FYI
@sleshchenko There is great interest in building multi-arch.

ericwill:
@bivasda1 so the build was successful, but please fix/update the following in a new PR:
* the `7.28.0-SNAPSHOT` tag shouldn't be there
* a nightly tag hasn't been published/updated -- likely this is what the `7.28.0-SNAPSHOT` tag should be

https://quay.io/repository/eclipse/che-devfile-registry?tab=tags

nickboldt:
There, I fixed it. :) https://quay.io/repository/eclipse/che-devfile-registry?tab=tags&tag=nightly

Also deleted the 7.28.0-SNAPSHOT tag

nickboldt:
But also... the image is not multiarch :(

```
#23 [linux/arm64 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
#23 sha256:3075bc83c5d59c32a9b1a25f5afd028722d2f71f259502bc813ddbcec8d90bf5
#23 0.092 /bin/sh: /tmp/install-editor-tooling.sh: not found
#23 ERROR: executor failed running [/bin/sh -c /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh]: exit code: 127

#12 [linux/amd64 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
#12 sha256:feb2a9698386ef81638fdeab39c0d81e0bbdad912bc578d7027b6a30d8144c40
#12 0.229 container process is already dead
#12 CANCELED

#17 [linux/ppc64le 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
#17 sha256:6d32b5832d93ea6889e549b5179b2c324f77d75fb4ce4be71c5176c98682b25b
#17 0.235 container process is already dead
#17 CANCELED
------
 > [linux/arm64 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh:
------
Dockerfile:10
--------------------
   8 |     # Install common terminal editors in container to aid development process
   9 |     COPY install-editor-tooling.sh /tmp
  10 | >>> RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
  11 |     
  12 |     USER 10001
--------------------
error: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh]: exit code: 127
Pushing quay.io/eclipse/che-antora-2.3:nightly to remote registry
An image does not exist locally with the tag: quay.io/eclipse/che-antora-2.3
The push refers to repository [quay.io/eclipse/che-antora-2.3]
Building quay.io/eclipse/che-cpp-rhel7:nightly based on registry.access.redhat.com/devtools/llvm-toolset-rhel7 ...
time="2021-03-08T15:15:42Z" level=warning msg="No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load"
```

-- https://github.com/eclipse/che-devfile-registry/runs/2057571221?check_suite_focus=true

and if I skopeo inspect I only get amd64:

```
$➔ skopeo --override-os ppc64le inspect docker://quay.io/eclipse/che-devfile-registry:nightly | jq -r .Architecture
amd64
```

Contrast with:

```
$➔ skopeo inspect docker://quay.io/eclipse/che-dashboard:multiarch-next --raw | jq -r '.manifests[].platform.architecture'
amd64
arm64
ppc64le
s390x
```

ArvinB:
@nickboldt I curious how this will get resolved, because this is similar to the issue I faced when trying to build registry images for `s390x`.

nickboldt:
well, there's the approach in https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml ...

.github/workflows/nightly-build-publish.yml
```diff
@@ -18,29 +18,55 @@ jobs:
...
-          docker push quay.io/eclipse/che-devfile-registry:nightly
-          docker tag che-devfile-registry quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
-          docker push quay.io/eclipse/che-devfile-registry:${SHORT_SHA1}
+      -
+        name: "Build and push"
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./build/dockerfiles/Dockerfile
+          platforms: ${{ steps.package.outputs.content }}
```
Prabhav-Thali, line 70: @nickboldt Image is still single arch because I guess slight change is required here. Instead of `platforms: ${{ steps.package.outputs.content }}`, it needs to be `platforms: ${{ steps.prep.outputs.platforms }}` as defined on line 37.

bivasda1:
> But also... the image is not multiarch :(
> 
> ```
> #23 [linux/arm64 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
> #23 sha256:3075bc83c5d59c32a9b1a25f5afd028722d2f71f259502bc813ddbcec8d90bf5
> #23 0.092 /bin/sh: /tmp/install-editor-tooling.sh: not found
> #23 ERROR: executor failed running [/bin/sh -c /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh]: exit code: 127
> 
> #12 [linux/amd64 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
> #12 sha256:feb2a9698386ef81638fdeab39c0d81e0bbdad912bc578d7027b6a30d8144c40
> #12 0.229 container process is already dead
> #12 CANCELED
> 
> #17 [linux/ppc64le 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
> #17 sha256:6d32b5832d93ea6889e549b5179b2c324f77d75fb4ce4be71c5176c98682b25b
> #17 0.235 container process is already dead
> #17 CANCELED
> ------
>  > [linux/arm64 5/6] RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh:
> ------
> Dockerfile:10
> --------------------
>    8 |     # Install common terminal editors in container to aid development process
>    9 |     COPY install-editor-tooling.sh /tmp
>   10 | >>> RUN /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh
>   11 |     
>   12 |     USER 10001
> --------------------
> error: failed to solve: rpc error: code = Unknown desc = executor failed running [/bin/sh -c /tmp/install-editor-tooling.sh && rm -f /tmp/install-editor-tooling.sh]: exit code: 127
> Pushing quay.io/eclipse/che-antora-2.3:nightly to remote registry
> An image does not exist locally with the tag: quay.io/eclipse/che-antora-2.3
> The push refers to repository [quay.io/eclipse/che-antora-2.3]
> Building quay.io/eclipse/che-cpp-rhel7:nightly based on registry.access.redhat.com/devtools/llvm-toolset-rhel7 ...
> time="2021-03-08T15:15:42Z" level=warning msg="No output specified for docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load"
> ```
> 
> -- https://github.com/eclipse/che-devfile-registry/runs/2057571221?check_suite_focus=true
> 
> and if I skopeo inspect I only get amd64:
> 
> ```
> $➔ skopeo --override-os ppc64le inspect docker://quay.io/eclipse/che-devfile-registry:nightly | jq -r .Architecture
> amd64
> ```
> 
> Contrast with:
> 
> ```
> $➔ skopeo inspect docker://quay.io/eclipse/che-dashboard:multiarch-next --raw | jq -r '.manifests[].platform.architecture'
> amd64
> arm64
> ppc64le
> s390x
> ```
@nickboldt @ericwill 
there are these images which are not multiarch. So do we need to excluded from build for releasing multiarch? 
```
che-antora-2.3          docker.io/antora/antora:2.3.3
che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
che-dotnet-3.1          mcr.microsoft.com/dotnet/core/sdk:3.1.301-buster
che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
che-php-7               quay.io/eclipse/che-php-base:7.4

che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7->not compatible for arm64
che-rust-1.39           rust:1.39.0-slim->not compatible for ppc64le
```

ericwill:
> @nickboldt @ericwill
> there are these images which are not multiarch. So do we need to excluded from build for releasing multiarch?
> 
> ```
> che-antora-2.3          docker.io/antora/antora:2.3.3
> che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk:2.2-stretch
> che-dotnet-3.1          mcr.microsoft.com/dotnet/core/sdk:3.1.301-buster
> che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
> che-php-7               quay.io/eclipse/che-php-base:7.4
> che-python-3.6          centos/python-36-centos7:1
> 
> che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7->not compatible for arm64
> che-rust-1.39           rust:1.39.0-slim->not compatible for ppc64le
> ```
> 
> Even we can enable multiarch for these repos.
> 
> ```
> che-php-7               quay.io/eclipse/che-php-base:7.4->registry.access.redhat.com/ubi8/php-72
> che-python-3.6          centos/python-36-centos7:1->registry.access.redhat.com/ubi8/python-36
> ```

Yes we need to exclude stacks that aren't available on multi-arch. For example if rust is x86 only, then it should be excluded from builds on other arches.

nickboldt:
For arch specific devfile exclusion filtering, see what we're dong downstream in CRW:

* if there's a .s390x or .ppc64le version of a devfile, use that instead of the default. If file is empty, remove the devfile entirely
  * https://github.com/redhat-developer/codeready-workspaces/blob/crw-2.7-rhel-8/dependencies/che-devfile-registry/build/scripts/swap_yamlfiles.sh
  * https://github.com/redhat-developer/codeready-workspaces/tree/crw-2.6-rhel-8/dependencies/che-devfile-registry/devfiles/00_java-microprofile-xp2 (x only)
  * https://github.com/redhat-developer/codeready-workspaces/tree/crw-2.6-rhel-8/dependencies/che-devfile-registry/devfiles/00_java-microprofile-xp1 (z & p only)

* switch from openjdk to openj9 image:
  *  https://github.com/redhat-developer/codeready-workspaces/blob/crw-2.7-rhel-8/dependencies/che-devfile-registry/build/scripts/swap_images.sh

* if the devfile references a sidecar which doesn't exist (or doesn't exist for the current arch), remove the devfile while adding digests:
  * https://github.com/redhat-developer/codeready-workspaces/blob/crw-2.7-rhel-8/dependencies/che-devfile-registry/build/scripts/write_image_digests.sh#L22-L26

bivasda1:
> For arch specific devfile exclusion filtering, see what we're dong downstream in CRW:
> 
> * if there's a .s390x or .ppc64le version of a devfile, use that instead of the default. If file is empty, remove the devfile entirely
>   
>   * https://github.com/redhat-developer/codeready-workspaces/blob/crw-2.7-rhel-8/dependencies/che-devfile-registry/build/scripts/swap_yamlfiles.sh
>   * https://github.com/redhat-developer/codeready-workspaces/tree/crw-2.6-rhel-8/dependencies/che-devfile-registry/devfiles/00_java-microprofile-xp2 (x only)
>   * https://github.com/redhat-developer/codeready-workspaces/tree/crw-2.6-rhel-8/dependencies/che-devfile-registry/devfiles/00_java-microprofile-xp1 (z & p only)
> * switch from openjdk to openj9 image:
>   
>   * https://github.com/redhat-developer/codeready-workspaces/blob/crw-2.7-rhel-8/dependencies/che-devfile-registry/build/scripts/swap_images.sh
> * if the devfile references a sidecar which doesn't exist (or doesn't exist for the current arch), remove the devfile while adding digests:
>   
>   * https://github.com/redhat-developer/codeready-workspaces/blob/crw-2.7-rhel-8/dependencies/che-devfile-registry/build/scripts/write_image_digests.sh#L22-L26

@nickboldt Thanks for your details :)

PR #346, opened 2021-03-17
Add s390x support in GH actions

Prabhav-Thali:
### What does this PR do?
- Implements [ci-multiarch.yaml](https://github.com/eclipse/che-dashboard/blob/master/.github/workflows/ci-multiarch.yaml) approach to build and push che-devfile-registry image.

This refers to [eclipse/che-devfile-registry#336 (comment)](https://github.com/eclipse/che-devfile-registry/pull/336#issuecomment-792789942) and also [18943](https://github.com/eclipse/che/issues/18943)

### DIFF
--- a/.github/workflows/nightly-build-publish.yml
+++ b/.github/workflows/nightly-build-publish.yml
@@ -14,12 +14,69 @@ on:
     branches: 
       - master
 
+env:
+  IMAGE: quay.io/eclipse/che-devfile-registry
+  VERSION: nightly
+
 jobs:
   nightly-build-publish:
     runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [amd64,arm64,ppc64le,s390x]
+    outputs:
+      amd64: ${{ steps.result.outputs.amd64 }}
+      arm64: ${{ steps.result.outputs.arm64 }}
+      ppc64le: ${{ steps.result.outputs.ppc64le }}
+      s390x: ${{ steps.result.outputs.s390x }}
     steps:
+      -
+        name: "Checkout source code"
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 1
+      -
+        name: "Set up QEMU"
+        uses: docker/setup-qemu-action@v1
+      -
+        name: "Set up Docker Buildx ${{ matrix.arch }}"
+        uses: docker/setup-buildx-action@v1
+      -
+        name: "Docker quay.io Login"
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_USERNAME }}
+          password: ${{ secrets.QUAY_PASSWORD }}
       - 
-        name: Clone source code
+        name: Login to docker.io
+        uses: docker/login-action@v1
+        with:
+          registry: docker.io
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      -
+        name: "Build and push ${{ matrix.arch }}"
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./build/dockerfiles/Dockerfile
+          platforms: linux/${{ matrix.arch }}
+          tags:  ${{ env.IMAGE }}:${{ env.VERSION }}-${{ matrix.arch }}
+          push: true 
+      -
+        id: result
+        name: "Build result outputs version"
+        if: ${{ success() }}
+        run: echo "::set-output name=${{ matrix.arch }}::${{ env.VERSION }}-${{ matrix.arch }}"
+
+  create-manifest:
+    needs: nightly-build-publish
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: "Checkout source code"
         uses: actions/checkout@v1
         with:
           fetch-depth: 1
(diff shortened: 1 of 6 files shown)


.github/workflows/nightly-build-publish.yml
```diff
@@ -14,34 +14,124 @@ on:
...
+      ppc64le: ${{ steps.result.outputs.ppc64le }}
+      s390x: ${{ steps.result.outputs.s390x }}
     steps:
-      - 
-        name: Clone source code
+      -
+        name: "Checkout source code"
         uses: actions/checkout@v1
         with:
-          fetch-depth: 1
```
ericwill, line 25: Any specific reason to change this?

.github/workflows/nightly-build-publish.yml
```diff
@@ -14,34 +14,124 @@ on:
...
+  create-manifest:
+    if: always()
+    needs: nightly-build-publish
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: "Checkout source code"
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 0
```
ericwill, line 77: Same comment about the `fetch-depth` here as above

.github/workflows/nightly-build-publish.yml
```diff
@@ -14,34 +14,124 @@ on:
...
+            echo "[!] The job 'nightly-build-publish' didn't provide any outputs. Can't create the manifest list."
+            exit 1;
+          fi
+          docker manifest create ${{ env.IMAGE }}:${{ env.VERSION }} $AMEND
+          docker manifest push ${{ env.IMAGE }}:${{ env.VERSION }}
+          docker manifest create ${{ env.IMAGE }}:${{ steps.prep.outputs.short_sha1 }} $AMEND
+          docker manifest push ${{ env.IMAGE }}:${{ steps.prep.outputs.short_sha1 }}
+
+  build-images:
+    if: always()
```
ericwill, line 121: Do we need this line if it's always going to happen?

.github/workflows/nightly-build-publish.yml
```diff
@@ -14,34 +14,124 @@ on:
...
+    needs: nightly-build-publish
+    runs-on: ubuntu-latest
+    steps:
+      - 
+        name: "Checkout source code"
+        uses: actions/checkout@v1
+        with:
+          fetch-depth: 0
```
ericwill, line 129: Same `fetch-depth` comment as above

.github/workflows/release.yml
```diff
@@ -39,22 +51,106 @@ jobs:
...
+        name: Prepare
+        id: prep
+        run: |
+          SHORT_SHA1=$(git rev-parse --short HEAD)
+          echo ::set-output name=short_sha1::${SHORT_SHA1}
+      -
+        name: "Docker quay.io Login"
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
```
ericwill, line 91: I believe this should be logging into docker.io, to prevent rate limiting

.github/workflows/nightly-build-publish.yml
Prabhav-Thali: No, kept release and nightly fetch-depth inline but I guess for nightly job we can have fetch-depth as 1 such that only one commit is fetched.

.github/workflows/nightly-build-publish.yml
Prabhav-Thali: In this case, it will always run after nightly-build-publish job has completed regardless of whether it is successful. Also just to confirm, should this be executed only when nightly-build-publish is successful or it can be executed as a separate standalone job?

.github/workflows/release.yml
Prabhav-Thali: As we are not performing push/pull to dockerhub, I had removed docker.io login. Should I add login to docker.io?

ericwill: Yes, I believe we need it as the build process will pull from dockerhub. We might run into rate limiting because of this, so it's important to login to docker.io

.github/workflows/release.yml
```diff
@@ -39,22 +51,106 @@ jobs:
...
+        name: "Set up QEMU"
+        uses: docker/setup-qemu-action@v1
+      -
+        name: "Set up Docker Buildx ${{ matrix.arch }}"
+        uses: docker/setup-buildx-action@v1
+      -
+        name: "Docker quay.io Login"
         uses: docker/login-action@v1
         with:
-          registry: docker.io
```
ericwill, line 46: We need to keep the docker.io login here as well

.github/workflows/nightly-build-publish.yml
```diff
@@ -14,12 +14,64 @@ on:
...
+      -
+        name: "Set up QEMU"
+        uses: docker/setup-qemu-action@v1
+      -
+        name: "Set up Docker Buildx ${{ matrix.arch }}"
+        uses: docker/setup-buildx-action@v1
+      -
+        name: "Docker quay.io Login"
+        uses: docker/login-action@v1
+        with:
```
ericwill, line 49: We should probably log in to docker.io here as well, just to be safe

.github/workflows/nightly-build-publish.yml
ericwill: If `nightly-build-publish` fails then building an image for that job wouldn't really serve much purpose, right? Am I understanding it correctly?

Prabhav-Thali: Yes, but build-images is building base images and happy path image. This should be executed only when nightly-build-publish(which builds che-devfile-registry image) job is successful? 

Also, to add more on this. Should we keep nightly build workflow similar to release workflow? wherein if build fails on 1 arch then further steps should not be executed.

.github/workflows/release.yml
Prabhav-Thali: okay, will add docker login back.

Prabhav-Thali:
@ericwill any update on this?

.github/workflows/nightly-build-publish.yml
ericwill: cc @nickboldt how do we handle this in other places? My instinct would say we should fail on any arch failure.

nickboldt: In CRW we *used* to have the build workflow scenario where if a multiarch build failed on one arch, we'd quietly remove those arches from the container.yaml used for Brew and just build it for the passing arches. This led to more problems than it was worth...
* PRO: could continue the overall build stack / release even when Z or P was broken
* PRO: if missing upstream artifacts, wouldn't waste time/resources/data centre heat building something we know was going to fail (or which used older artifacts and would therefore lead to unexpected containter content [eg., theia container 2.8 built from theia tarballs from 2.7]; this is NOT a use case for Che since Che doesn't do previously-bootstrapped offline builds, and will always fetch latest from GH and npm and go registries.

* CON: Z Dev/QE could be lost at sea wondering where their updated containers were, and why only x and P got updated in quay
* CON: might end up with a case where we can't deploy on Z or P because the operator-metadata payload was incomplete (eg., if theia or traefik failed to build)

Do we run automated tests on all arches (on push or for PRs)? can those tests be skipped if the arch-specific containers failed to build? Or do we want to gate a given push/PR from being released if an arch is missing? 

I'd say the answer may depend on flakiness of the arch-specific builds. If z fails to build 1 time in 50, then its absence should be a bug that would prevent the release. Buf it if fails 1 time in 3, then maybe it's worth skipping over it if it's going to hinder progress 66.6% of the time.

ericwill:
I've just reverted part of #336 due to missing dev container/happy path images.

Adding multi-arch build support needs to be done correctly, this is the second time we've had to go back due to some failure. Here's what's needed:
* What Nick mentioned in this thread: https://github.com/eclipse-che/che-devfile-registry/pull/346#discussion_r605663190
    * Likely we should fail on build failure, and if it starts happening often then keep an eye on it
* Exclusion system:
    * Not all devfiles work on all arches, and not all dev containers are available on all arches
    * We need a way to only build the images we need for a given arch

Prabhav-Thali:
>   * Likely we should fail on build failure, and if it starts happening often then keep an eye on it

Sure, I'll make the necessary changes to implement this and update the PR.

> * Exclusion system:
>
>   * Not all devfiles work on all arches, and not all dev containers are available on all arches
>   * We need a way to only build the images we need for a given arch

Hi @bivasda1, Are you working on this change? Will you be raising PR soon for this? In that case, @ericwill can we separate the exclusion functionality in another PR and keep this for che-devfile-registry image?

bivasda1:
> > * Likely we should fail on build failure, and if it starts happening often then keep an eye on it
> 
> Sure, I'll make the necessary changes to implement this and update the PR.
> 
> > * Exclusion system:
> >   
> >   * Not all devfiles work on all arches, and not all dev containers are available on all arches
> >   * We need a way to only build the images we need for a given arch
> 
> Hi @bivasda1, Are you working on this change? Will you be raising PR soon for this? In that case, @ericwill can we separate the exclusion functionality in another PR and keep this for che-devfile-registry image?

@ericwill  `We need a way to only build the images we need for a given arch`->I am working on, Facing few issues, will update you soon for this.

nickboldt:
This won't be merged, as we're now going to use Travis for s390x.

PR #352, opened 2021-03-22
Cache images in browser cache with etags

sleshchenko:
### What does this PR do?
This PR makes browser cache devfile images with ETags.

I tested it and image is reloaded after change.
The time benefit is not much, but it presents.
Screenshot_20210322_102043

### What are alternatives?
Better alternative could be if we hard cache images for some period like 1 week, and make browser reload updated resources by renaming images files to contains their hashes.
But it requires more changes on devfile registry.

https://github.com/eclipse/che/issues/16296

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -5,7 +5,7 @@ DirectoryIndex README.md
 AddType image/svg+xml svg svgz
 AddEncoding gzip svgz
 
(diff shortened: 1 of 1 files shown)


sleshchenko:
When I tested if with `fast 3g mode`, it does not seem to work. I assume more headers are needed.

PR #364, opened 2021-04-06
feature(disk cleanup) use --rm flag to purge images from local docker registry after pushing to remote

nickboldt:
### What does this PR do?

use --rm flag to purge images from local docker registry after they're pushed to remote

Change-Id: Ibf077cba35048b3bc72b7d4025eef5ffbd0f191c

### Screenshot/screencast of this PR
N/A

last week we ran out of disk building the che-devfile-registry's stack/sidecar images. This seeks to improve that problem by deleting images from the local registry after we've pushed them to quay.

If building locally is a common usecase (outside of GH actions) we might want to refactor things a bit so that the `--rm` flag is an optional one passed to make-release.sh, rather than being a default setting when make-release.sh calls into the `arbitrary-users-patch` scripts. 

### How to test this PR?
N/A

### PR Checklist

[As the author of this Pull Request I made sure that:](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#pull-request-template-and-its-checklist)

- [x] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [x] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [x] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [x] [The repository devfile is up to date and works, if one exists](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [x] [Sections What issues does this PR fix or reference and How to test this PR completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)
- [x] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers

Reviewers, please comment how you tested the PR when approving it.

### DIFF
--- a/arbitrary-users-patch/build_images.sh
+++ b/arbitrary-users-patch/build_images.sh
@@ -23,8 +23,13 @@ TAG=${TAG:-${DEFAULT_TAG}}
 NAME_FORMAT="${REGISTRY}/${ORGANIZATION}"
 
 PUSH_IMAGES=false
-if [ "$1" == "--push" ]; then
+if [ "$1" == "--push" ] || [ "$2" == "--push" ]; then
   PUSH_IMAGES=true
+fi
+
+RM_IMAGES=false
+if [ "$1" == "--rm" ] || [ "$2" == "--rm" ]; then
+  RM_IMAGES=true
 fi
 
 BUILT_IMAGES=""
@@ -34,8 +39,12 @@ while read -r line; do
   echo "Building ${NAME_FORMAT}/${base_image_name}:${TAG} based on $base_image ..."
   docker build -t "${NAME_FORMAT}/${base_image_name}:${TAG}" --no-cache --build-arg FROM_IMAGE="$base_image" "${SCRIPT_DIR}"/ | cat
   if ${PUSH_IMAGES}; then
-    echo "Pushing ${NAME_FORMAT}/${base_image_name}:${TAG}" to remote registry
+    echo "Pushing ${NAME_FORMAT}/${base_image_name}:${TAG} to remote registry"
     docker push "${NAME_FORMAT}/${base_image_name}:${TAG}" | cat
+  fi
+  if ${RM_IMAGES}; then # save disk space by deleting the image we just published
+    echo "Deleting ${NAME_FORMAT}/${base_image_name}:${TAG} from local registry"
+    docker rmi "${NAME_FORMAT}/${base_image_name}:${TAG}"
   fi
   BUILT_IMAGES="${BUILT_IMAGES}    ${NAME_FORMAT}/${base_image_name}:${TAG}\n"
 done < "${SCRIPT_DIR}"/base_images
--- a/.github/workflows/nightly-build-publish.yml
+++ b/.github/workflows/nightly-build-publish.yml
@@ -59,10 +59,10 @@ jobs:
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
       - name: Build and push base images
-        run: ./arbitrary-users-patch/build_images.sh --push
+        run: ./arbitrary-users-patch/build_images.sh --push --rm
 
       - name: Build and push happy path image
-        run: ./arbitrary-users-patch/happy-path/build_happy_path_image.sh --push
+        run: ./arbitrary-users-patch/happy-path/build_happy_path_image.sh --push --rm
       -
         name: "Build and push"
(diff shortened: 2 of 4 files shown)


arbitrary-users-patch/build_images.sh
```diff
@@ -23,20 +23,29 @@ TAG=${TAG:-${DEFAULT_TAG}}
...
+if [ "$1" == "--rm" ] || [ "$2" == "--rm" ]; then
+  RM_IMAGES=true
+fi
+
 BUILT_IMAGES=""
 while read -r line; do
   base_image_name=$(echo "$line" | tr -s ' ' | cut -f 1 -d ' ')
   base_image=$(echo "$line" | tr -s ' ' | cut -f 2 -d ' ')
   echo "Building ${NAME_FORMAT}/${base_image_name}:${TAG} based on $base_image ..."
   docker build -t "${NAME_FORMAT}/${base_image_name}:${TAG}" --no-cache --build-arg FROM_IMAGE="$base_image" "${SCRIPT_DIR}"/ | cat
```
benoitf, line 40: what if we use docker buildx ? AFAIK there is no local registry publication in that case

mkuznyetsov (approved):

arbitrary-users-patch/build_images.sh
mkuznyetsov: maybe when multiarch build will be implemented again, the one that works that is, we shall see if this option will be needed.

benoitf: buildx can be used for a single arch

nickboldt: that's a problem for IBM PEs to solve :) 

perhaps we have to run each container build in a separate thread / action step / container so that they're not running out of disk. We barely have enough disk for the sidecars for x64 + 3-4 arches of the devfile registry, so when we're building the dozen sidecars x 3-4 arches we'll definitely be out of disk super fast.

nickboldt: "the tool" being a GH action instead of the shell script that can be run locally? 

Yeah, that's great until you're stuck waiting for a GH action to fire for 5hrs (che-theia) and can't build locally. :) 

But conceptually, yes, I agree. reusing GH actions instead of writing our own wrappers is a noble cause. As long as it solves our dogfooding/development workflows.

ericwill: +1 for using buildx, we will eventually switch to that for multi-arch anyway

benoitf: > "the tool" being a GH action instead of the shell script that can be run locally?

 no, I mean the tool = `docker buildx` not `docker build` (it can run locally as well)
just that build, publish, delete, better caching options is already managed, no need to add new .sh scripts around that

nickboldt: > IMHO it looks cleaner to rely on the tools than do .sh wrappers to handle publish, remove, etc when the tool can handle both publish and remove by itself

Right, but then when we want to call the same set of tool commands for both nightly and release builds, we can either:
* use the shell script wrappers, or
* copy the same docker commands boilerplate into two github action scripts.

Why not collect the commands into a single shell script for portability? Then you can run it:
* as a nightly
* as a release
* locally

ericwill (approved):
Approving as we can always move to `buildx` later, this will solve our problem for tomorrow's release.

nickboldt:
Thx, merged and cherrypicked to 7.28.x as a8d819b

PR #368, opened 2021-04-07
Devfile registry general maintenance

ericwill:

### What does this PR do?
* Use digests in `base_images` list
* Build dev container images as part of PR check
* Use ubuntu-20.04 as a runner instead of latest

### Screenshot/screencast of this PR

N/A

No issue, just general fixups.

### How to test this PR?

* Run `arbitrary-users-patch/build_images.sh` and observe that digests are used instead of tags.
* Open a PR against this branch and observe that the PR check now builds dev container images as well

### PR Checklist

[As the author of this Pull Request I made sure that:](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#pull-request-template-and-its-checklist)

- [x] [The Eclipse Contributor Agreement is valid](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-eclipse-contributor-agreement-is-valid)
- [x] [Code produced is complete](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-produced-is-complete)
- [x] [Code builds without errors](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#code-builds-without-errors)
- [ ] [Tests are covering the bugfix](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#tests-are-covering-the-bugfix)
- [x] [The repository devfile is up to date and works](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#the-repository-devfile-is-up-to-date-and-works)
- [x] [Sections `What issues does this PR fix or reference` and `How to test this PR` completed](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#sections-what-issues-does-this-pr-fix-or-reference-and-how-to-test-this-pr-completed)
- [x] [Relevant user documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [Relevant contributing documentation updated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#relevant-contributing-documentation-updated)
- [x] [CI/CD changes implemented, documented and communicated](https://github.com/eclipse/che/blob/master/CONTRIBUTING.md#cicd-changes-implemented-documented-and-communicated)

### Reviewers

Reviewers, please comment how you tested the PR when approving it.


arbitrary-users-patch/base_images
```diff
@@ -1,18 +1,17 @@
...
-che-nodejs10-community  node:10.16
-che-nodejs12-community  node:12.18
-che-nodejs10-ubi        registry.access.redhat.com/ubi8/nodejs-10
-che-nodejs8-centos      registry.centos.org/che-stacks/centos-nodejs
-che-php-7               quay.io/eclipse/che-php-base:7.4
-che-python-3.6          centos/python-36-centos7:1
-che-python-3.8          python:3.8.6-slim
-che-quarkus             quay.io/quarkus/centos-quarkus-maven:21.0-java11
-che-rust-1.39           rust:1.39.0-slim
+che-antora-2.3          docker.io/antora/antora@sha256:fbf31b3be4e5dd215c82f9b5b403c5e746f68bcf50ced9b7d41b6d1414aa5714
```
benoitf, line 1: do we need a script that fetches digest and update this file ? Or like a .lock file with tag/digest mapping
Because it looks like it will be really hard to maintain these digests ( like to know the original tag alias for the digest, check for updates, etc.)

nickboldt: seems like we'd want something akin to what's in the related images of the che operator CSVs...
* given a list of images:tags, 
* generate a list of images@sha256:digests (using `skopeo` and `jq`), but 
* retain the source file w/ the tags so we can 
* fetch a new list of digests at intervals. 

Then we can:
* stay current but also stay locked down, and be able to 
* roll back or manually replace a tag w/ a digest if we know latest is broken and we want something older

ericwill: I don't think a script will solve the problem, the issue we had yesterday was that the one of the tags was overwritten and it broke the build. +1 for a file that shows the mappings though

arbitrary-users-patch/base_images
```diff
@@ -1,18 +1,17 @@
...
+che-cpp-rhel7           registry.access.redhat.com/devtools/llvm-toolset-rhel7@sha256:24573aefd172be551ab4ae047a31f5ba5cd59f2e5bc207188f3b7045ac9d7c72
+che-dotnet-2.2          mcr.microsoft.com/dotnet/core/sdk@sha256:db9b38d066fdbdd5ac5ce862076c27e4ae17f2b57cbacd03bdc3b24e06696ae8
+che-dotnet-3.1          mcr.microsoft.com/dotnet/core/sdk@sha256:c87fd21bdebb2c2d573ecd703981476e5b8ac6a0e96d134722bb672e1c231fea
+che-golang-1.12         golang@sha256:8b0f4fee390ca1308ef5c522accad9f7d98a4134fcde1b2cff33df1cbb1b3b97
+che-golang-1.14         golang@sha256:a60bc0c6be7e352336044e68db7089d8604864933fcbc0d7cb2285d60b0b2d98
+che-java11-gradle       gradle@sha256:c25f15608cd9bee8e290dd5654f9a784bea4505719fdedd0ad7ac640bed0fc50
+che-java11-maven        maven@sha256:1d29ccf46ef2a5e64f7de3d79a63f9bcffb4dc56be0ae3daed5ca5542b38aa2d
+che-java8-maven         maven@sha256:69b40237b342fee9bf996b81110f4dab250a4bc6a2ee52866965101eda066324
+che-nodejs10-community  node@sha256:f4b6f471cdd4b66b27eef899e7f8423ecd9fbfc863b2cb7a59978a7f64c8e0c3
```
nickboldt, line 10: personal pet peeve: why don't we mention the registry here? should be docker.io/... not just image name. That way you'll never accidentally get the wrong image if it happens to be on multiple registries and your docker config file picks the wrong registry first.

.github/workflows/pr-checks.yml
```diff
@@ -26,17 +26,26 @@ jobs:
...
 
-  build:
+  build-dev-containers:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v1
+
+      - name: Build dev container images
+        run: ./arbitrary-users-patch/build_images.sh --rm
```
nickboldt, line 44: if you remove them but don't publish them, how will the subsequent step be able to resolve them as base image for the happy path image build?

arbitrary-users-patch/base_images
ericwill: Good point :+1:

.github/workflows/pr-checks.yml
ericwill: If we publish them then every PR run will overwrite the `nightly` tag for all the dev containers. Also in this case we aren't building the HP image, only the dev containers and the main devfile registry image. It's different than the release job/script.

nickboldt: yep, sounds good.

arbitrary-users-patch/base_images
nickboldt: +1, new 3-column format rules.

nickboldt (approved):
image

benoitf (approved):
on a side note we should get rid of nodejs 8 and 10 images

ericwill:
> on a side note we should get rid of nodejs 8 and 10 images

Agreed, looks like all the images could benefit from being checked and updated also.

PR #473, opened 2021-09-29
[update] Update digests in base_images

vitaliy-guliy:


vitaliy-guliy:
It was a test PR, ignore it.

PR #663, opened 2022-07-05
Added a v2 devfile and a dogfooding link

l0rd:
### What does this PR do?

There are 3 commits:

- The first commit adds a version 2 devfile and a badge to start in dogfooding cluster
- The second commit removes the samples that don't have a v2 devfile
- The third commit removes the v1 devfiles


svor:
Here is the issue https://github.com/eclipse/che/issues/21119 which describes the changes

svor (approved):
I'm OK with these changes. Just a few comments:
1. I think we can remove an old devfile from the root folder
2. **3-publish** command fails because che-devfile-registry-prod component doesn't have **surge** and **tree** commands
screenshot-che-dogfooding apps che-dev x6e0 p1 openshiftapps com-2022 07 14-16_51_26

But it could be improved in this issue: https://github.com/eclipse/che/issues/19305

.vscode/extensions.json
```diff
@@ -0,0 +1,12 @@
...
+    // for the documentation about the extensions.json format
+    "recommendations": [
+      "redhat.vscode-xml",
+      "redhat.vscode-yaml",
+      "eamodio.gitlens",
+      "donjayamanne.githistory",
+      "timonwong.shellcheck",
+      "GitHub.vscode-pull-request-github"
+    ]
+  }
```
svor, line 12: New line is missing

README.md
```diff
@@ -1,6 +1,7 @@
 ![Next Build Status](https://github.com/eclipse/che-devfile-registry/actions/workflows/next-build.yml/badge.svg)
 ![Release Build Status](https://github.com/eclipse/che-devfile-registry/actions/workflows/release.yml/badge.svg)
 [![Contribute](https://www.eclipse.org/che/contribute.svg)](https://workspaces.openshift.com#https://github.com/eclipse/che-devfile-registry)
+[![Contribute (nightly)](https://img.shields.io/static/v1?label=nightly%20Che&message=for%20maintainers&logo=eclipseche&color=FDB940&labelColor=525C86)](https://che-dogfooding.apps.che-dev.x6e0.p1.openshiftapps.com/#https://github.com/eclipse-che/che-devfile-registry?df=.devfile-v2.yaml&che-editor=che-incubator/che-code/insiders)
```
svor, line 4: since we have che-incubator/che-code/insiders in che-editor.yaml and dogfooding instance uses che-code by default we can avoid this parameter in the factory url

.vscode/extensions.json
l0rd: 👍

README.md
l0rd: 👍

l0rd:
@svor thanks for the review. I have addressed the points you mentioned except the removal of the old devfile.yaml as that would mean that we would break opening the repository on developer sandbox since it hasn't been upgraded to Dev Spaces yet.

For the things that should be improved in a separate PR and in the context of #19305:

1. Fix the `build/dev/build-devfile-registry.sh` so that it generates dw and dwt (using `./generate_devworkspace_templates.sh`)
2. Fix the UDI image with some [tweaks that allow using buildah in rootless mode](https://github.com/l0rd/tilt-example-java/blob/main/Dockerfile.dev#L9)
3. Add a `build image` command that builds the devfile registry image using buildah
4. Add a `publish image` command that push the locally built image in a container registry using buildah 
5. Add a `run with local image` command that runs a custom devfile registry image using the locally built image
6. Add a `replace Che devfile registry` command that patch the `CheCluster` to use the locally built devfile registry image

svor (approved):

PR #799, opened 2023-11-16
fix:set fixed version for devfile/api

svor:
### What does this PR do?
Set fixed version (2.2.1-alpha-1667236163) of devfile/api

Fixes build of devworkspace generator


akurinnoy (approved):
LGTM

